PR #3, opened 2017-04-27
General improvements

vassyz:
- fixes the address field
- adds the email field
- adds image URL property
- adds an integration example

I have a few questions, maybe @pgreci would be the best one to answer:

- Do I have to add versioning to Go packages? It doesn't seem to have one. Should I use the GitHub Releases feature?
- Is there a way of ignoring the example folder when a user does a `go get` on this package?
- I added the `URL` property to the `Image` struct, but I'm not sure what the others are used for: `Data` is in Byte format, and `Type` is a string that is not being used anywhere. The other SDKs return the selfie as a base64 embedded image.


vassyz:
Took @pgreci advice into account and set image URL as a method. It uses octet-stream instead of base64. I also removed the unnecessary copy call on the data object.

PR #9, opened 2017-10-26
 [SDK-143]: Saved user selfie, and then display alongside the Base64URI

echarrod:


example/main.go
```diff
@@ -32,8 +36,12 @@ func profile(w http.ResponseWriter, r *http.Request) {
 
 	if err == nil {
 		templateVars := map[string]interface{}{
-			"profile": profile,
-			"selfie":  template.URL(profile.Selfie.URL())}
+			"profile":         profile,
+			"selfiebase64URL": template.URL(profile.Selfie.URL())}
```
vassyz: selfiebase64URL -> selfieBase64URL

echarrod: Good spot

vassyz:
Not really related to this task, but could you add something like `log.Printf("About to listen on 8080")` before `http.ListenAndServe(":8080", nil)`. Also, in the documentation, you can mention opening the localhost:8080 page after `* start the server go run main.go`.

echarrod:
Ok I can do. In the README it says 
> The callback URL will be: http://your-local-url:8080/profile.

a few lines above "start the server go run main.go". It seems a bit repetitive to mention this again? But I guess it's better to do that than to cause potential confusion

vassyz:
If you find it repetitive, let's mention this in the log string: `About to listen on 8080. Go to https://localhost:8080/`

README.md
```diff
@@ -167,17 +167,18 @@ The examples also uses the `YOTI_APPLICATION_ID` environment variable to display
...
-        * [X] Photo `selfie`
-        * [X] Given Names `given_names`
-        * [X] Family Name `family_name`
-        * [X] Mobile Number `phone_number`
-        * [X] Email address `email_address`
-        * [X] Date of Birth `date_of_birth`
-        * [X] Address `postal_address`
-        * [X] Gender `gender`
-        * [X] Nationality `nationality`
+        * [X] User ID `ID`
```
vassyz: There are some unnecessary tabs at the end of this line.

example/main.go
```diff
@@ -56,6 +57,7 @@ func main() {
 	http.HandleFunc("/profile", profile)
 	http.Handle("/images/", http.StripPrefix("/images",
 		http.FileServer(http.Dir(path.Join(rootdir, "images/")))))
+	log.Printf("About to listen and serve on 8080. Go to http://your-local-url:8080/")
```
vassyz: Can we have `localhost` or `127.0.0.1` instead of `your-local-url`? I know the docs are being more generic, but I'd rather be more explicit in the example. People can change the code anyway. Once we have this, this is good to merge.

PR #11, opened 2018-02-07
[SDK-259]: Added 'Full Name' and 'AgeVerified' attributes

echarrod:


vassyz:
Any particular reason why you added full name underneath family name and not before given names, similar to the other SDKs?

echarrod:
Do you mean in the README? 
I did it that way in [Python](https://github.com/getyoti/yoti-python-sdk/commit/88da24c3d2a149f1b43b02d4b0e338d0f8ccc5d3), and it was done that way in [PHP](https://github.com/getyoti/yoti-php-sdk/commit/ae1d13f5234d04a4a23cc523cde7869aa4bf07ee) too

PR #12, opened 2018-03-16
[SDK-251]: AML checking functionality

echarrod:
- AML check functionality which involved allowing POST http requests, including a payload/content to the message
- Added example to demonstrate this new functionality
- Tidied up code


README.md
```diff
@@ -142,6 +145,74 @@ You can run the unit tests for this project by executing the following command i
...
+        "On Fraud list: %s",
+        strconv.FormatBool(result.OnFraudList)))
+log.Printf(
+    fmt.Sprintf(
+        "On Watch list: %s",
+        strconv.FormatBool(result.OnWatchList)))
+}
+```
+
+Additionally an [example AML application](/amlexample/main.go) is provided in the examples folder. As with the other example project, you'll need to rename `.env.example` to `env` and fill in the variables within the file.
```
vassyz: `Additionally,` and `.env`

vassyz:
Can you group the two examples into one folder? Something similar to `examples/aml` and `examples/profile`

.vscode/launch.json
```diff
@@ -4,17 +4,24 @@
     // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
```
leonard-kramer: Do we really want .vscode in the project? It feels like something that should be removed in .gitignore..

echarrod: We looked at this a little while ago and thought it would be useful for devs running the example project so they wouldn't have to set anything up, whilst also not polluting the repo too much. 

Do you think IDE-specific files should always be excluded?

amlexample/main.go
```diff
@@ -0,0 +1,58 @@
...
+	"os"
+	"strconv"
+
+	"github.com/getyoti/yoti-go-sdk"
+
+	_ "github.com/joho/godotenv/autoload"
+)
+
+var sdkID = os.Getenv("YOTI_CLIENT_SDK_ID")
+var key, _ = ioutil.ReadFile(os.Getenv("YOTI_KEY_FILE_PATH"))
```
leonard-kramer: What happens if this fails?

amlexample/main.go
```diff
@@ -0,0 +1,58 @@
...
+
+	result, err := client.PerformAmlCheck(amlProfile)
+
+	if err != nil {
+		log.Printf(
+			fmt.Sprintf(
+				"Unable to retrieve AML result. Error: %s", err))
+	} else {
+		log.Printf(
+			fmt.Sprintf(
```
leonard-kramer: You don't need fmt.Sprintf here, Printf is able to format (that's the f in Printf)

amlexample/main.go
```diff
@@ -0,0 +1,58 @@
...
+				"On PEP list: %s",
+				strconv.FormatBool(result.OnPEPList)))
+		log.Printf(
+			fmt.Sprintf(
+				"On Fraud list: %s",
+				strconv.FormatBool(result.OnFraudList)))
+		log.Printf(
+			fmt.Sprintf(
```
leonard-kramer: As above

httprequester.go
```diff
@@ -34,15 +46,15 @@ func doRequest(uri string, headers map[string]string) (result *httpResponse, err
 
 	defer resp.Body.Close()
 
-	var body []byte
-	if body, err = ioutil.ReadAll(resp.Body); err != nil {
+	var responseBody []byte
+	if responseBody, err = ioutil.ReadAll(resp.Body); err != nil {
 		return
```
leonard-kramer: Maybe log the error?

.vscode/launch.json
leonard-kramer: Yes, it's better to have instructions in the README on how to set it up in different IDEs (if there is any difficulty). The problem with having ide specific files is that not everybody runs the same IDE, and it's possible that a different IDE might clash with these. Also these files will be useless clutter to a sizable percentage of users of the SDK.

image.go
```diff
@@ -2,31 +2,37 @@ package yoti
 
 import "encoding/base64"
 
-type ImageType int
+type imageType int
```
leonard-kramer: Since you're exporting the values below I think you should export the type as well (and document it).

yotiattributevalue.go
```diff
@@ -1,43 +1,47 @@
 package yoti
 
-type AttributeType int
+type attributeType int
```
leonard-kramer: Should be exported.

yoticlient.go
```diff
@@ -208,17 +197,74 @@ func decryptCurrentUserReceipt(receipt *receiptDO, key *rsa.PrivateKey) (result
...
 
+// PerformAmlCheck performs an Anti Money Laundering Check (AML) for a particular user.
+// Returns three boolean values: 'OnPEPList', 'OnWatchList' and 'OnFraudList'.
+func (client *Client) PerformAmlCheck(amlProfile AmlProfile) (AmlResult, error) {
+	return performAmlCheck(amlProfile, doRequest, client.SdkID, client.Key)
+}
+
+func performAmlCheck(amlProfile AmlProfile, requester httpRequester, sdkID string, keyBytes []byte) (result AmlResult, err error) {
+	var key *rsa.PrivateKey
+	var httpMethod = "POST"
```
leonard-kramer: Why don't you make these a series of const values, i.e httpMethodPOST, httpMethodGET, etc.

then when you call createHeaders further down you can just use httpMethodPOST, easy to figure out what it is withouth having to scroll up, plus keeping values like this as const is good practice.

yoticlient.go
```diff
@@ -233,6 +279,26 @@ func getTimestamp() string {
...
+	}
+
+	var authDigest string
+	if authDigest, err = getAuthDigest(endpoint, key, httpMethod, content); err != nil {
+		return
+	}
+
+	headers = make(map[string]string)
+
+	headers["X-Yoti-Auth-Key"] = authKey
```
leonard-kramer: The values in here should be const with comments.

yoticlient.go
```diff
@@ -233,6 +279,26 @@ func getTimestamp() string {
...
+	headers["X-Yoti-Auth-Digest"] = authDigest
```
leonard-kramer: Same

yoticlient.go
```diff
@@ -233,6 +279,26 @@ func getTimestamp() string {
...
+	headers["X-Yoti-SDK"] = sdkIdentifier
```
leonard-kramer: Same

.vscode/launch.json
echarrod: Would you feel it would be ok if `.vscode` lived in the /examples folder instead? That seems like less clutter, and would still operate in the same way

leonard-kramer: I'm a bit confused why it's so important to get it in? Are there any particular difficulty getting vs code to work?

echarrod: It just means that if you you try to debug from anywhere but the file you're wanting to run, it will show this error and create the `launch.json` file and `.vscode` folder for you if you click `open launch.json`
image
Where as if you have the folder and file, it gives you dropdowns for the different run configurations from anywhere.

echarrod: @leonard-kramer ok Seb made a good suggestion - we can add a `CONTRIBUTING.md` which can have this launch.json file in a gist. And I can `git rm` the folder, and add it to `.gitignore`, sound ok?

leonard-kramer: yeah, that sounds like a better way.

CONTRIBUTING.md
```diff
@@ -0,0 +1,43 @@
...
+
+## Commit Process
+
+1) `go fmt` to format the code
+1) `go vet` reports suspicious constructs
+1) `go test` to run the tests
+
+## VS Code
+
+For developing in VS code, use the following `launch.json` file (placed inside a `.vscode` folder) to easily run the examples from VS Code:
```
vassyz: `VS Code` with uppercase `C`.

CONTRIBUTING.md
```diff
@@ -0,0 +1,43 @@
...
+
+```json
```
vassyz: Because you're using comments, it's probably better to set the syntax to `javascript`. Seems like the comments look all read otherwise: https://github.com/getyoti/yoti-go-sdk/blob/SDK-251-AML/CONTRIBUTING.md

README.md
```diff
@@ -142,14 +145,81 @@ You can run the unit tests for this project by executing the following command i
...
+
+amlProfile := yoti.AmlProfile{
+    GivenNames: givenNames,
+    FamilyName: familyName,
+    Address:    amlAddress}
+
+result, err := client.PerformAmlCheck(amlProfile)
+
+log.Printf(
+    fmt.Sprintf(
```
leonard-kramer: These examples should be updated to not use fmt.

examples/aml/main.go
```diff
@@ -0,0 +1,59 @@
...
+	"os"
+	"strconv"
+
+	"github.com/getyoti/yoti-go-sdk"
+
+	_ "github.com/joho/godotenv/autoload"
+)
+
+var sdkID = os.Getenv("YOTI_CLIENT_SDK_ID")
+var key, err = ioutil.ReadFile(os.Getenv("YOTI_KEY_FILE_PATH"))
```
leonard-kramer: I would do this as 
```
var (
  sdkID string
  key ioutil.File
  client *yoti.Client
)
```
and then initialise in main()
That way you can catch and handle the error

examples/aml/main.go
```diff
@@ -0,0 +1,59 @@
...
+
+var client = yoti.Client{
+	SdkID: sdkID,
+	Key:   key}
+
+func main() {
+	if err != nil {
+		log.Printf("Unable to retrieve `YOTI_KEY_FILE_PATH`. Error: `%s`", err)
+
```
leonard-kramer: Don't think there needs to be a blank new line here.

examples/profile/main.go
```diff
@@ -19,7 +19,7 @@ import (
 
 var sdkID = os.Getenv("YOTI_CLIENT_SDK_ID")
 var key, _ = ioutil.ReadFile(os.Getenv("YOTI_KEY_FILE_PATH"))
```
leonard-kramer: See my comment to the same above.

PR #13, opened 2018-04-05
[SDK-284]: Added IsAgeVerified helper

echarrod:


yoti_test.go
```diff
@@ -333,6 +333,53 @@ func TestYotiClient_PerformAmlCheck_Unsuccessful(t *testing.T) {
...
+	if *isAgeVerified != true {
+		t.Error("Expected true")
+		return
+	}
+
+	return
+}
+
+func TestYotiClient_ParseIsAgeVerifiedValue_False(t *testing.T) {
+	trueValue := []byte("false")
```
vassyz: Shouldn't this be `falseValue`?

yotiuserprofile.go
```diff
@@ -30,6 +30,9 @@ type UserProfile struct {
 	// DateOfBirth represents the user's date of birth. This will be nil if not provided by Yoti
 	DateOfBirth *time.Time
 
+	// IsAgeVerified represents the result of the age verification check on the user. The bool will be true if they passed, false if they failed, and nil if there was no check
+	IsAgeVerified *bool
```
LKDraytonBrown: How come a pointer to the bool? (and time.Time?)

echarrod: I originally wanted it to be a pointer so that the state that there wasn't an age check performed could be represented with a nil pointer. As it turns out, this isn't necessary with the latest control flow - as it will never be set if this is the case, and will remain nil. So i will change to just a bool

echarrod: RE *time.Time, I believe this will have been implemented this way when the SDK was originally created, is it problematic?

LKDraytonBrown: If its not set it will be: false, this is the default bool value (https://tour.golang.org/basics/12).
So this will not be ok if you are tracking three states. Guess its ok (although I think an enum better models 3 states).

echarrod: As discussed, having the nil value is required for the case where the AgeVerified check is not carried out, so I will leave it as a pointer

PR #18, opened 2018-07-04
SDK-376: Self-signed certificate

echarrod:
- Added self-signed certificate to example project
- Cert and private key files are generated if not present (and ignored in source control)
- Redirection from HTTP->HTTPS

### DIFF
--- a/examples/profile/certificatehelper.go
+++ b/examples/profile/certificatehelper.go
@@ -50,34 +50,27 @@ func pemBlockForKey(priv interface{}) *pem.Block {
 	}
 }
 
-func certificatePresenceCheck(certPath string, keyPath string) error {
+func certificatePresenceCheck(certPath string, keyPath string) (present bool) {
 	if _, err := os.Stat(certPath); os.IsNotExist(err) {
-		return err
-	} else if _, err := os.Stat(keyPath); os.IsNotExist(err) {
-		return err
+		return false
 	}
-	return nil
+	if _, err := os.Stat(keyPath); os.IsNotExist(err) {
+		return false
+	}
+	return true
 }
 
-func generateSelfSignedCertificate(certPath string, keyPath string, host string) error {
-	var priv interface{}
-	var err error
-
-	priv, err = rsa.GenerateKey(rand.Reader, rsaBits)
+func generateSelfSignedCertificate(certPath, keyPath, host string) error {
+	priv, err := rsa.GenerateKey(rand.Reader, rsaBits)
 	if err != nil {
 		log.Printf("failed to generate private key: %s", err)
 		return err
 	}
 
-	var notBefore time.Time
-	if len(validFrom) == 0 {
-		notBefore = time.Now()
-	} else {
-		notBefore, err = time.Parse("Jan 2 15:04:05 2006", validFrom)
-		if err != nil {
-			fmt.Fprintf(os.Stderr, "Failed to parse creation date: %s\n", err)
-			return err
-		}
+	notBefore, err := parseNotBefore(validFrom)
+	if err != nil {
+		log.Printf("failed to parse 'Not Before' value of cert using validFrom %q, error was: %s", validFrom, err)
+		return err
 	}
 
 	notAfter := notBefore.Add(validFor)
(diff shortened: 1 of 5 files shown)


examples/profile/main.go
```diff
@@ -64,13 +65,45 @@ func profile(w http.ResponseWriter, r *http.Request) {
...
 
 func main() {
-	rootdir, _ := os.Getwd()
+	// Check if the cert files are available.
+	err := certificatePresenceCheck("yotiSelfSignedCert.pem", "yotiSelfSignedKey.pem")
+	// If they are not available, generate new ones.
+	if err != nil {
+		err = generateSelfSignedCertificate("yotiSelfSignedCert.pem", "yotiSelfSignedKey.pem", "127.0.0.1:8080")
+		if err != nil {
+			log.Fatal("Error: Couldn't create https certs.")
```
bucky-boy: I don't know Go, but.... will this try to continue executing, after it's done a log.Fatal?

echarrod: No it won't continue - it's in effect a call to `Print()` followed by a call to `os.Exit(1)`.

examples/profile/certificatehelper.go
```diff
@@ -0,0 +1,143 @@
...
+		return err
+	}
+
+	certOut, err := os.Create(certPath)
+	if err != nil {
+		log.Printf("failed to open "+certPath+" for writing: %s", err)
+		return err
+	}
+	pem.Encode(certOut, &pem.Block{Type: "CERTIFICATE", Bytes: derBytes})
+	certOut.Close()
```
pn-santos: probably better to extract to method and use `defer certOut.Close()` inside that method:
```
err := createPemFile(certPath, derBytes)
if err != nil {
    log.Printf("failed to create pem file at %q: %s", certPath, err)
    return err
}
```

examples/profile/certificatehelper.go
```diff
@@ -0,0 +1,143 @@
...
+	log.Print("written yotiSelfSignedCert.pem\n")
+
+	keyOut, err := os.OpenFile(keyPath, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0600)
+	if err != nil {
+		log.Print("failed to open "+keyPath+" for writing:", err)
+		return err
+	}
+	pem.Encode(keyOut, pemBlockForKey(priv))
+	keyOut.Close()
```
pn-santos: same here (extract to method and use `defer keyOut.Close()`)

examples/profile/main.go
```diff
@@ -64,13 +65,45 @@ func profile(w http.ResponseWriter, r *http.Request) {
 }
 
 func main() {
-	rootdir, _ := os.Getwd()
+	// Check if the cert files are available.
+	err := certificatePresenceCheck("yotiSelfSignedCert.pem", "yotiSelfSignedKey.pem")
```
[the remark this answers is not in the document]
echarrod: I will make them constants. I think it would be a good idea to not make them configurable, since they are included in the .gitignore file

examples/profile/main.go
```diff
@@ -64,13 +65,45 @@ func profile(w http.ResponseWriter, r *http.Request) {
...
+
+	log.Printf("About to configure HTTPS redirection")
+	configureHTTPSRedirection()
+
+	log.Printf("About to listen and serve on 8080. Go to https://localhost:8080/")
+	http.ListenAndServeTLS(":8080", "yotiSelfSignedCert.pem", "yotiSelfSignedKey.pem", nil)
+}
+
+func configureHTTPSRedirection() {
+	go func() {
```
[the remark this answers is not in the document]
echarrod: Ok, makes sense

examples/profile/certificatehelper.go
```diff
@@ -0,0 +1,143 @@
...
+		Subject: pkix.Name{
+			Organization: []string{"Yoti"},
+		},
+		NotBefore: notBefore,
+		NotAfter:  notAfter,
+
+		KeyUsage:              x509.KeyUsageKeyEncipherment | x509.KeyUsageDigitalSignature,
+		ExtKeyUsage:           []x509.ExtKeyUsage{x509.ExtKeyUsageServerAuth},
+		BasicConstraintsValid: true,
+	}
```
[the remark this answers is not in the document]
pn-santos: humm ok nvm then I look at another piece of code that didn't have the values aligned but it must have been because the values were wrapped over more than one line

examples/profile/certificatehelper.go
```diff
@@ -122,22 +115,60 @@ func generateSelfSignedCertificate(certPath string, keyPath string, host string)
...
+	log.Printf("written %s\n", keyPath)
+
+	return nil
+}
+
+func createPemFile(certPath string, derBytes []byte) error {
+
 	certOut, err := os.Create(certPath)
+	defer certOut.Close()
```
pn-santos: you should place this after the nil check (before `pem.Encode`), no point in closing when it wasn't opened (because of an error)

examples/profile/certificatehelper.go
```diff
@@ -122,22 +115,60 @@ func generateSelfSignedCertificate(certPath string, keyPath string, host string)
...
 	pem.Encode(certOut, &pem.Block{Type: "CERTIFICATE", Bytes: derBytes})
-	certOut.Close()
-	log.Print("written yotiSelfSignedCert.pem\n")
 
+	return nil
+}
+
+func createKeyFile(keyPath string, privateKey interface{}) error {
 	keyOut, err := os.OpenFile(keyPath, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0600)
+	defer keyOut.Close()
```
pn-santos: same here

examples/profile/main.go
```diff
@@ -86,24 +89,29 @@ func main() {
...
+func redirectHandler(w http.ResponseWriter, req *http.Request) {
 	hostParts := strings.Split(req.Host, ":")
-	http.Redirect(w, req, "https://"+hostParts[0]+req.RequestURI, http.StatusMovedPermanently)
+	http.Redirect(
+		w,
+		req,
+		fmt.Sprintf(
+			"https://%s%s",
+			hostParts[0],
+			req.RequestURI),
```
pn-santos: you could probably keep the `Sprintf` all in one line no?

PR #20, opened 2018-07-23
SDK-342: Using formatted address for address

echarrod:
- Added feature to take `formatted address` from `structured postal address` if `address` is missing
- Made a few changes based on `gometalinter`, there are lots more suggestions, I've added ticket SDK-449 to look at the rest of these

### DIFF
--- a/yoti_test.go
+++ b/yoti_test.go
@@ -26,7 +26,7 @@ func TestYotiClient_KeyLoad_Failure(t *testing.T) {
 
 	if err == nil {
 		t.Error("Expected failure")
-	} else if strings.HasPrefix(err.Error(), "Invalid Key") == false {
+	} else if !strings.HasPrefix(err.Error(), "Invalid Key") {
 		t.Errorf("expected outcome type starting with %q instead received %q", "Invalid Key", err.Error())
 	}
 }
@@ -159,7 +159,7 @@ func TestYotiClient_ParseProfile_Success(t *testing.T) {
 	dob := time.Date(1980, time.January, 1, 0, 0, 0, 0, time.UTC)
 	if profile.DateOfBirth == nil {
 		t.Error(`expected date of birth but it was not present in the returned profile`)
-	} else if profile.DateOfBirth.Equal(dob) == false {
+	} else if !profile.DateOfBirth.Equal(dob) {
 		t.Errorf("expected date of birth %q instead received %q", profile.DateOfBirth.Format(time.UnixDate), dob.Format(time.UnixDate))
 	}
 }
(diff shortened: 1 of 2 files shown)


yoti_test.go
```diff
@@ -368,6 +368,106 @@ func TestYotiClient_UnmarshallJSONValue_ValidValue(t *testing.T) {
...
+
+	var structuredAddressBytes = []byte(`[
+	{		
+		"address_format": 2,
+		"building": "House No.86-A",
+		"formatted_address": "` + formattedAddressText + `"
+	}
+	]`)
+	structuredAddress, err := unmarshallJSON(structuredAddressBytes)
+
```
pn-santos: remove newline (so err check is next to assignment) - same below

yoticlient.go
```diff
@@ -171,6 +172,10 @@ func getActivityDetails(requester httpRequester, encryptedToken, sdkID string, k
 					}
 				}
 			}
+			formattedAddress, err := getFormattedAddressIfAddressIsMissing(result)
+			if formattedAddress != "" && err == nil {
+				result.Address = formattedAddress
+			}
```
pn-santos: what if err != nil ?

yoticlient.go
```diff
@@ -171,6 +181,12 @@ func getActivityDetails(requester httpRequester, encryptedToken, sdkID string, k
 					}
 				}
 			}
+			formattedAddress, err := getFormattedAddressIfAddressIsMissing(result)
+			if err == nil {
+				result.Address = formattedAddress
```
sidibos: Does this mean if both `structured_postal_address.formatted_address` and `postal_address` are set. It will always return `formatted_address`?

Or this logic should be used directly on line `138`

echarrod: Good spot, I removed that logic in the last commit as I was trying to trim it down, didn't think of that. 

Can't put it on line 138 as there's no guarantee that the structured_postal_address would have been process by then, but I'll change this logic here

yoticlient.go
```diff
@@ -131,6 +132,8 @@ func getActivityDetails(requester httpRequester, encryptedToken, sdkID string, k
 					parsedTime, err := time.Parse("2006-01-02", string(attribute.Value))
 					if err == nil {
 						result.DateOfBirth = &parsedTime
+					} else {
+						log.Printf("Unable to parse `date_of_birth` value: %q. Error: %q", attribute.Value, err)
 					}
 				case "postal_address":
```
sidibos: On another note, why are we using `switch/case` on profile attribute retrieval? Imo we need some architecture here, otherwise this SDK will be very hard to extend/maintain at some points.

echarrod: Yes, switch/case is not ideal. The architecture here is something we need to address definitely, but is probably beyond the scope of this PR

PR #22, opened 2018-09-03
SDK-488: Profile Processing

echarrod:
I've attempted to break this up as much as I can, so there will be this PR and one for adding Anchors, and potentially another for updating the Example project.

### New Profile object 
There is now a new profile object which is now retrieved through an `ActivityDetails` object. Because of the addition of `otherPartyProfile` in future, it is called `userProfile` on this, but the underlying object is Profile, to avoid clashing with the old profile method (maintained for backwards compatibility). 

This new profile has attributes which are parsed based on their type (from Protobuf).

### For the Generics alternative:
There are `AttributeXXX.` classes for each type of attribute, e.g. `AttributeString` and `AttributeImage`.
These classes each have distinct extension methods which mean that if, for example, `Nationality` is an `AttributeString`, then you can call Profile.Nationality().String(), but you can'd do this for non-string attribute types. 
This means the codebase has to have one of these classes for each type, but I see this as being the best way to achieve a generics alternative.

### Other
- attrtype_string.go is a generated file which allows you to call .String() on the (enum-type) class AttrType, to get it's string representation. This functionality is not included by default in Go, but this provides a fairly elegant solution.
- I wasn't able to partition it completely, so there are a few references to anchors in places in this PR, this will make sense after that has been checked in
- There are tests for every new piece of functionality, these will be checked in in the next PR, it's very difficult to split them up since they're all in one class (ticket created to organise the test files)
- See the README for an overview of how it all fits together


README.md
```diff
@@ -132,7 +151,7 @@ if err == nil {
 }
 ```
 
-Where `yourUserSearchFunction` is a piece of logic in your app that is supposed to find a user, given a userID.
+Where `yourUserSearchFunction` is a piece of logic in your app that is supposed to find a user, given a RememberMeID.
```
bucky-boy: _rememberMeId_, or _userId_?
Looks like we've referred to this as _userId_ elsewhere.  I don't why there would be a difference, or why we can't just stick to one name.

Can anyone shed any light?

If it was _userId_ before, maybe we should just leave it.

README.md
```diff
@@ -228,18 +247,18 @@ Visiting `https://localhost:8080/` should show a Yoti Connect button
...
     * [X] Selfie `Selfie`
-    * [X] Selfie URL `Selfie.URL`
+    * [X] Selfie URL `Selfie().Image`
     * [X] Given Names `GivenNames`
     * [X] Family Name `FamilyName`
     * [X] Full Name `FullName`
     * [X] Mobile Number `MobileNumber`
     * [X] Email Address `EmailAddress`
-    * [X] Age / Date of Birth `DateOfBirth`
-    * [X] Age / Is Age Verified `IsAgeVerified`
```
bucky-boy: Is this the _IsAgeVerified_ that returns a boolean?

attributejson.go
```diff
@@ -0,0 +1,56 @@
...
+
+//AttributeJSON is a Yoti attribute which returns a string as its value
+type AttributeJSON struct {
+	attribute
+}
+
+func newAttributeJSON(byteValue []byte, anchors []Anchor, name string, attrType AttrType) (result AttributeJSON) {
+	if attrType != AttrTypeJSON {
+		log.Printf("Cannot create JSON attribute with non-JSON type: %q", attrType.String())
+		return
```
pn-santos: Again not really sure this in inline with go "best practices", usually if a function has the possibility of not doing what it's supposed to (due to bad input or transient error), it should "signal" that through it's interface by returning `err`. One of the "things" in go is to make error handling mandatory (by returning `err` in function where stuff can go wrong).

crypto.go
```diff
@@ -120,3 +124,28 @@ func generateNonce() (string, error) {
 
 	return uuid, nil
 }
+
+func decryptToken(encryptedConnectToken string, key *rsa.PrivateKey) (result string, err error) {
+	// token was encoded as a urlsafe base64 so it can be transfered in a url
+	var cipherBytes []byte
+	if cipherBytes, err = urlSafeBase64ToBytes(encryptedConnectToken); err != nil {
+		return
```
pn-santos: I believe it's usually preferable to do `return nil, err` than to take advantage of the implicit returns, it makes the code more obvious and easier to figure what what is actually be returned (but this is just my perception of it, not sure if it's correct so take with a grain of salt eheh)

attributegeneric.go
```diff
@@ -0,0 +1,41 @@
+package yoti
+
+//AttributeGeneric is a Yoti attribute which returns a generic value, which must be explicitly cast
+type AttributeGeneric struct {
```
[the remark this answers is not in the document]
echarrod: I do agree that GenericAttribute is better, but since all the files are under the same directory, I thought it would be useful if they were grouped together alphabetically, I think it makes it a little more organised. If we think that's not important at all then I can change them

README.md
```diff
@@ -98,17 +98,36 @@ Keeping your settings and access keys outside your repository is highly recommen
...
-Before you inspect the user profile, you might want to check whether the user validation was successful.
-This is done as follows:
+You can then get the following values from the activityDetails struct:
 
 ```Go
-profile, err := client.GetUserProfile(yotiOneTimeUseToken)
-if err != nil {
-  // handle unhappy path
-}
+rememberMeID := activityDetails.RememberMeID
```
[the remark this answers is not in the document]
bucky-boy: We now have SDK-558 for this question.
This is so simple... we need a decision so we can get this done.

attributeimage.go
```diff
@@ -0,0 +1,82 @@
...
+		log.Printf("Unable to create base64 URL for type: %q", ai.Type.String())
+		return
+	}
+
+	return "data:image/" + imageType + ";base64;," + base64EncodedImage
+}
+
+// Image returns the value of an attribute in the form of a Yoti Image object
+func (ai AttributeImage) Image() *Image {
+	var image *Image
```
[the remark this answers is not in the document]
echarrod: If we don't have this here, it would mean that the user retrieving the selfie wouldn't be like `selfie := userProfile.Selfie().Image()` (see the README). Although I do agree it's not entirely neccesary, I think it's good for it to be retrieved in the same way as the other attribute values, i.e. with an extension method with a name which matches the type it will return.

attributeimage.go
```diff
@@ -0,0 +1,82 @@
...
+	switch ai.Type {
+	case AttrTypeJPEG:
+		image = &Image{
+			Type: ImageTypeJpeg,
+			Data: ai.Value}
```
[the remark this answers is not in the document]
echarrod: I've actually replaced this in another change, as we don't need the imageType here, as we can just use AttrType, will check in with your other suggestions

attributeimage.go
```diff
@@ -0,0 +1,82 @@
...
+	case AttrTypePNG:
+		image = &Image{
+			Type: ImageTypePng,
+			Data: ai.Value}
+	default:
+		log.Printf("Unable to parse Image value of type: %q", ai.Type.String())
```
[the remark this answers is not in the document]
echarrod: As above, the changes I will check in will supercede this

attributejson.go
echarrod: In the SDKs we've taken the stance to not fail the attribute parsing process if an error occurs, so that the other attributes can still be processed, instead we are just logging the error. Given this, would returning an err at this stage provide us with any other benefits?

crypto.go
echarrod: That makes sense to me, but that will have to be `return "", err` because you can't use nil for the string return type, right?

attributegeneric.go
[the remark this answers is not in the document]
pn-santos: You can do that organization with a `attribute` directory, and placing all different attribute types in there.

attributejson.go
pn-santos: The question is who/what decides to continue on, presumably there is a "main" method that is parsing all the attributes. The further down the stack, the less opinionated the code should be in regards to "business" logic and error handling.

What I mean is the lower layers should return errors and then at some point  there will be a "layer" that is controlling the main flow of a certain action where all "business" rules will be will be enforced and will decide whether to abort or to continue on in case of error (or in case of a particular error). Even the logging decision, usually you either log or you propagate an error, but not both.

I'm not too into design decisions here so consider this comment only food for though :wink: but from a design perspective it makes for cleaner code if there is a single error handling approach: always propagate err, until you can't or until you reach code that "owns" the business behavior (and enforces it).

attributegeneric.go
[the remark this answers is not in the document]
echarrod: I can do that in every other language I know of, but as I understand it it's not possible to have sub-directories which are in the same package in Go, as Go maps package paths onto directories (non-recursively). It's not entirely clear from the [official Go guidelines](https://blog.golang.org/package-names) ("Package Paths" section), but it's mentioned on the [Package Clause](https://golang.org/ref/spec#Package_clause) of the Go spec. 

So I don't think putting them in a separate directory is an option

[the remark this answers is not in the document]
pn-santos: But why does stuff need to be in the same package? if organizationally it feels like they should all be "related" and in the same dir, it probably also means from an API pov that they should probably be in their own package and export whatever methods they need to export.

attributegeneric.go
```diff
@@ -0,0 +1,41 @@
...
+		anchors: anchors,
+		name:    name,
+	}
+
+	ag.Type = attrType
+	ag.Value = byteValue
+
+	return AttributeGeneric{
+		attribute: ag,
+	}
```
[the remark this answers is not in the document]
pn-santos: > is the principle that we should return a pointer if the struct is reasonably large?

Yes that's the "official" guidance, when you have a struct with two or three fields then return the struct, more than that return a pointer.

echarrod:
Last commit is a large restructure for the attribute parsing process. We wait until the last possible moment before casting the attribute to the desired type. This removes the need for casting to an intermediary type and averts the need for generics. 
The exception for this is the generic .GetAttribute function. By default this is returning a string, and the "data" field can be retrieved too, which is bytes. Clients can then cast to their desired type if they wish.

DateOfBirth and StructuredPostalAddress can encounter errors during this casting process. I have opted to include these as a field on the struct, rather than returning the error alongside the struct, so that it doesn't break up the process of calling profile.DateOfBirth().Value, .Anchors etc., but I am open to changing this. 

The second commit is just moving the attribute files to a separate package. Shouldn't need to look at the file content itself, but rather just check the package separations are logical.

yoticlient.go
```diff
@@ -37,30 +43,55 @@ type Client struct {
...
-func getActivityDetails(requester httpRequester, encryptedToken, sdkID string, keyBytes []byte) (result UserProfile, err error) {
+// GetActivityDetails requests information about a Yoti user using the one time use token generated by the Yoti login process.
+// It returns the outcome of the request. If the request was successful it will include the users details, otherwise
+// it will specify a reason the request failed.
+func (client *Client) GetActivityDetails(token string) (ActivityDetails, []string) {
+	_, activityDetails, errStrings := getActivityDetails(doRequest, token, client.SdkID, client.Key)
+	return activityDetails, errStrings
+}
+
+func getActivityDetails(requester httpRequester, encryptedToken, sdkID string, keyBytes []byte) (userProfile UserProfile, activityDetails ActivityDetails, errStrings []string) {
```
echarrod: @LKDraytonBrown & @pn-santos - we are thinking of removing this list of error strings returned (and also from the TimeAtttribute and JSONAttribute), and instead just log the errors, and return nil for the attributes that cannot be parsed. I know this is not the traditional Go way, in that we want to propagate any errors, but we don't want clients to fail to get any of the other attributes if there is a problem with one. 

Do you think this is an acceptable solution to not return the errors?

echarrod: Drayton suggested to tag @serverside, who might be able to answer this question

README.md
echarrod: As discussed, this will be RememberMeID everywhere now - [SDK-572](https://lampkicking.atlassian.net/browse/SDK-572)

README.md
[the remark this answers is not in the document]
echarrod: As discussed, this will be RememberMeID everywhere now - [SDK-572](https://lampkicking.atlassian.net/browse/SDK-572)

yoticlient.go
pn-santos: So a `nil` attribute means there was an error processing that attribute?

If that is so, then I guess that is acceptable, though if there is a way for us to "categorize" the errors, it would be preferable to return a "partial success" data structure in which you have the ability to describe/identify the error for a particular attribute (even if for now there is only 1 error type). 

Returning `nil` (i.e. the equivalent of a "there was _an_ error processing attribute X") gives the client the ability to know there was an error, but that's it. Maybe that's enough, but consider whether it might be of use to the client to know the type of error (either now or in a hypothetical future). The `nil` approach will not allow you to "extend" the error information without a backwards incompatible change (if it ever becomes a necessity).

attribute/genericattribute.go
```diff
@@ -0,0 +1,25 @@
+package attribute
+
+import (
+	"github.com/getyoti/yoti-go-sdk/anchor"
+)
+
+//GenericAttribute is a Yoti attribute which returns a generic value
+type GenericAttribute struct {
```
LKDraytonBrown: Including the pkg name in the pkg member names results in stutter. How about just 'Generic', since the pkg name prefix lets us know its a attribute?

attribute/genericattribute.go
```diff
@@ -0,0 +1,25 @@
...
+	Name    string
+	Value   string
+	Data    []byte
+	Type    AttrType
+	Anchors []*anchor.Anchor
+}
+
+//NewGenericAttribute creates a new generic attribute
+func NewGenericAttribute(a *Attribute) *GenericAttribute {
```
LKDraytonBrown: stutter, how about 'NewGeneric'?

attribute/imageattribute.go
```diff
@@ -0,0 +1,25 @@
+package attribute
+
+import (
+	"github.com/getyoti/yoti-go-sdk/anchor"
+)
+
+//ImageAttribute is a Yoti attribute which returns an image as its value
+type ImageAttribute struct {
```
LKDraytonBrown: stutter... the fix in this case is to create an image sub pkg... it is common for a go pkg to be light weight.

attribute/jsonattribute.go
```diff
@@ -0,0 +1,46 @@
...
+func UnmarshallJSON(byteValue []byte) (result interface{}, err error) {
+	var unmarshalledJSON interface{}
+	err = json.Unmarshal(byteValue, &unmarshalledJSON)
+
+	if err != nil {
+		return nil, err
+	}
+
+	return unmarshalledJSON, err
+}
```
LKDraytonBrown: I'm not sure this function give you much. How about making it a method of the JSONAttribute type that accepts an interface to unmarshal too? Something like:
```
func (j *JSONAttribute) Unmarshal(v *interface{}){
...
err = json.Unmarshal(j.byteValue, v)
...
}

File: attribute/image.go [hunk 16]
```diff
@@ -0,0 +1,52 @@
+package attribute
+
+import (
+	"encoding/base64"
+)
+
+//Deprecated: Will be removed in v3.0.0 - use AttrType instead. ImageType Image format
+type ImageType int
```
LKDraytonBrown: You seem to be duplicating AttrType?

attribute/timeattribute.go
```diff
@@ -0,0 +1,33 @@
...
+	Anchors []*anchor.Anchor
+	Err     error
+}
+
+//NewTimeAttribute creates a new Time attribute
+func NewTimeAttribute(a *Attribute) *TimeAttribute {
+	parsedTime, err := time.Parse("2006-01-02", string(a.Value))
+	if err != nil {
+		log.Printf("Unable to parse time value of: %q. Error: %q", a.Value, err)
+	}
```
LKDraytonBrown: Ignoring this error could result in a corrupt TimeAttribute

attribute/timeattribute.go
```diff
@@ -0,0 +1,33 @@
...
+
+	return &TimeAttribute{
+		Name:    a.Name,
+		Value:   &parsedTime,
```
LKDraytonBrown: Wont this result in a NPE if there has been an error parsing?

attribute/yotiattribute.go
```diff
@@ -0,0 +1,78 @@
...
+)
+
+// Attribute represents a small piece of information about a Yoti user such as a photo of the user or the
+// user's date of birth.
+type Attribute struct {
+	Name    string
+	Value   []byte
+	Type    AttrType
+	Anchors []*anchor.Anchor
+}
```
LKDraytonBrown: After some thought, I think using the actual attribute proto is the better bet.

attribute/yotiattribute.go
```diff
@@ -0,0 +1,78 @@
...
+		result = "image/png"
+
+	case AttrTypeJSON:
+		result = "application/json; charset=UTF-8"
+
+	default:
+		log.Printf("Unable to find a matching MIME type for value type %q", attributeType.String())
+	}
+	return
+}
```
LKDraytonBrown: Why not define this on the type?
```   
func (a AttrType) GetMIMEType() (result string)

File: conversion.go [hunk 21]
```diff
@@ -33,3 +39,48 @@ func base64ToBytes(base64Str string) ([]byte, error) {
...
+		attrType = attribute.AttrTypeInterface
+	}
+
+	return &attribute.Attribute{
+		Name:    protoAttribute.Name,
+		Value:   protoAttribute.Value,
+		Type:    attrType,
+		Anchors: processedAnchors,
+	}
+}
```
LKDraytonBrown: What is the value of this re-typing?

yotiprofile.go
```diff
@@ -0,0 +1,146 @@
...
+
+// GivenNames represents the user's given names. Will be nil if not provided by Yoti
+func (p Profile) GivenNames() *attribute.StringAttribute {
+	for _, a := range p.AttributeSlice {
+		if a.Name == attrConstGivenNames {
+			return attribute.NewStringAttribute(a)
+		}
+	}
+	return nil
+}
```
LKDraytonBrown: This seems to imply that a profile has only a single given name. This may not be true as a user can register more than a single document each possibly resulting in more than a single given_names attribtues

attribute/imageattribute.go
LKDraytonBrown: Actually ignore that comment, think there is a better way...

attribute/image.go
echarrod: This method was there previously, I had removed it in an earlier commit, before realising the best course of action would be to mark it as deprecated in this release, and then remove it in the next

attribute/yotiattribute.go
echarrod: It would be good to not create a wrapper attribute, but it does make it a lot harder to test. If I have this wrapper, I can create attributes in a clear way in the tests, and test they perform as expected.

attribute/jsonattribute.go
echarrod: At the point that we are unmarshalling it, it is of type "*Attribute", not "*JSONAttribute". So I could add this as an extension method to "Attribute", but I think it would make more sense to have it here.

attribute/yotiattribute.go
echarrod: We're calling this from attribute.image, so I could add it there as well. But I thought it might be used in other places in future, and so might be better to place it in the general yoti(attribute) class.

conversion.go
echarrod: Were you thinking that we should just keep the protobuf types throughout, and not convert them? I guess it provides us with a cleaner attribute object, which doesn't have the extra fields which the protobuf attributes have.

yotiprofile.go
echarrod: None of the SDKs are set up to have multiple attributes with the same name. We can discuss this with Quirino.

attribute/time.go
```diff
@@ -0,0 +1,34 @@
...
+type Time struct {
+	Name    string
+	Value   *time.Time
+	Type    AttrType
+	Anchors []*anchor.Anchor
+	Err     error
+}
+
+//NewTime creates a new Time attribute
+func NewTime(a *Attribute) *Time {
```
echarrod: @LKDraytonBrown (and @pn-santos?), was hoping to get your thoughts on this method - we thought about having the return type as (time, err) for this method, but this breaks up the attribute retrieval process, e.g. we can't have:
```golang
emailAddress := userProfile.EmailAddress().Value
dateOfBirth := userProfile.DateOfBirth().Value
address := userProfile.Address().Value
```
we instead need to have:
```golang
emailAddress := userProfile.EmailAddress().Value
dob, err := userProfile.DateOfBirth()
if err != nil {
    //handle error
} else {
    dateofBirth = dob
}
address := userProfile.Address().Value
```

The same is true for the StructuredPostalAddress attribute.

So we can either:
1. Return (time, err), and check for the error
1. Return time, but also have an `err` field on the struct, as is done here

I would prefer the latter, but I know this would mean that it's not obvious to clients that they need to check for this error.
I would appreciate your thoughts on whether this would be suitable from a Go perspective.

pn-santos: The go dev will probably say option 1 (return time, err). Errors always are explicit to "force" the developer/client to deal with them.

Looking at the `Time` struct though, it doesn't make sense for it to have an `Err` field, error of what? An `err` is the result of some _operation_ (in go lingo, function), however that struct seems to represent only "data". Having an `err` seems to imply it represents the result of some operation but that's it's name (Time) is communicating.

Which least to the the method used: `DateOfBirth()`. This seems like an accessor method, i.e. "getting" a value/data. Why would you get an error obtaining a piece of data? If the method is actually _doing_ something (other than just returning a stored value) that can result in an error I think then the name name should reflect what it's _doing_, it should be the name of the _operation_ and not simply the name of the data in case of "success".

echarrod: This is a by product of having to use a generics alternative: 
It is indeed an accessor method, but we are having to wait until the last moment to convert the "Yoti(Attribute)" to a "Time(Attribute)", so it is not possible in the current format to retrieve it before this point. 

So you're saying it would be more appropriate to call them `GetDateOfBirth` and `GetStructuredPostalAddress` (or something similar), and return an error alongside? Or are these also accessor names that would suggest it wouldn't return an error, and it should be `ParseDateOfBirth` and `ParseStructuredPostalAddress`?

yoticlient.go
```diff
@@ -199,10 +175,124 @@ func getActivityDetails(requester httpRequester, encryptedToken, sdkID string, k
 		}
 	}
 
+	return userProfile, activityDetails, errStrings
```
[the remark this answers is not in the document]
echarrod: Yep compiles, the signature goes off the page to the right, was it cutting this off for you? Or there's the external method GetActivityDetails, were you looking at that?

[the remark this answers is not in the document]
echarrod: Maybe I missed it out of an earlier commit

yoticlient.go
echarrod: This is a by product of having to use a generics alternative: 
It is indeed an accessor method, but we are having to wait until the last moment to convert the "Yoti(Attribute)" to a "Time(Attribute)", so it is not possible in the current format to retrieve it before this point. 

So you're saying it would be more appropriate to call them `GetDateOfBirth` and `GetStructuredPostalAddress` (or something similar), and return an error alongside? Or are these also accessor names that would suggest it wouldn't return an error, and it should be `ParseDateOfBirth` and `ParseStructuredPostalAddress`?

attribute/time.go
pn-santos: I'm saying looking at it from a go-perspective, any function than can return an error should return `err` alongside the data it's supposed to return in a non-failure scenario.

Regarding the function names, I was trying to say that the method name should reflect what it does. If I understand you correctly you are just converting types (casting) and the error is only when that operation fails is that right? (which is unlikely to happen right?) 

If the "operation" being done is just casting from one type to the other then nvm what I said. Leaving them as-is (e.g. `DateOfBirth()`) is fine.

tl;dr I'd propose returning `(time, err)`

attribute/generic.go
```diff
@@ -0,0 +1,25 @@
+package attribute
+
+import (
+	"github.com/getyoti/yoti-go-sdk/anchor"
+)
+
+//Generic is a Yoti attribute which returns a generic value
+type Generic struct {
+	Name    string
+	Value   string
```
LKDraytonBrown: Do you really want all your attributes to have a string representation? This may get difficult for non-string values... ie. a selfie pic.

attribute/yoti.go
```diff
@@ -0,0 +1,79 @@
...
+		result = "image/png"
+
+	case AttrTypeJSON:
+		result = "application/json; charset=UTF-8"
+
+	default:
+		log.Printf("Unable to find a matching MIME type for value type %q", attributeType.String())
+	}
+	return
+}
```
LKDraytonBrown: What are you gaining with this file? I really dont think this gives you much more than code to maintain. This GetMIMEType function is probably the most useful part, maybe put it in a map?

conversion.go
LKDraytonBrown: Yes, I think that using the proto throughout will mean less to maintain and no data/fidelity loss in translation. I initially suggested embedding it, but I think just have use an internal field.

yotiprofile.go
LKDraytonBrown: Yes, I think this is worth discussing. It maybe that they have done something similar to the client, which is to have a priority order here and choose the top priority to return... we would still support multiple right up to this point I think.

yotiprofile.go
```diff
@@ -0,0 +1,146 @@
...
+	attrConstGivenNames              = "given_names"
+	attrConstFamilyName              = "family_name"
+	attrConstFullName                = "full_name"
+	attrConstMobileNumber            = "phone_number"
+	attrConstEmailAddress            = "email_address"
+	attrConstDateOfBirth             = "date_of_birth"
+	attrConstAddress                 = "postal_address"
+	attrConstStructuredPostalAddress = "structured_postal_address"
+	attrConstGender                  = "gender"
+	attrConstNationality             = "nationality"
```
LKDraytonBrown: These are part of the core-api... not sure redefining them is worth it/a good idea...

conversion.go
echarrod: I've created [SDK-624](https://lampkicking.atlassian.net/browse/SDK-624) to look at this, but am going to keep it like this for now.

yotiprofile.go
echarrod: This is something which is likely to be implemented in the SDKs in the future. Since it will require quite a large overhaul for each one, we won't be doing it right now.

attribute/generic.go
echarrod: This Generic type is only returned for the GetAttribute function. We are ensuring we make each of the SDKs work in the same way as much as possible, using the Java SDK as a base.
 
Although we considered only having a byte value, the Java SDK returns a string value, and so we are making this work in the same way, as changing it in Java would require a breaking change. 

It might produce strange results for image types, but it also gives a convenience method for what most of the attributes are - string types.

attribute/yoti.go
echarrod: I think I have removed this intermediary type at one point of this work, and it made testing a lot more difficult - having to create these protobuf attributes instead of a new type. 

Additionally if we had the protobuf attribute, we'd have the raw anchors on this. Since we don't want to expose these to the clients, this would mean we'd have to parse these anchors to extract the information we need at the point of the user accessing the attribute from the profile. I believe we want to be doing this before this stage.

echarrod: Where would we make the map for MimeType if we were to make one? I think an extension method is quite appropriate for its use

yotiprofile.go
```diff
@@ -2,20 +2,21 @@ package yoti
...
-	attrConstFamilyName              = "family_name"
-	attrConstFullName                = "full_name"
-	attrConstMobileNumber            = "phone_number"
-	attrConstEmailAddress            = "email_address"
-	attrConstDateOfBirth             = "date_of_birth"
-	attrConstAddress                 = "postal_address"
-	attrConstStructuredPostalAddress = "structured_postal_address"
-	attrConstGender                  = "gender"
-	attrConstNationality             = "nationality"
+	attrConstSelfie                  = attributenames.Selfie
```
LKDraytonBrown: Why rem-map them? Why not just use them as provided from the attributenames pkg?

echarrod: It was originally because they were referenced in the tests too, but I will just change them there too to use them straight from the attributenames pkg

LKDraytonBrown: Yeah, should be able to rename them in place to the attributenames.<> and then once done should be able to delete this. Should result in more maintainable code in the long run. (one source of truth)

attribute/generic.go
LKDraytonBrown: I think the "make each of the SDKs work in the same way as much as possible" could make things really difficult if you try too hard... I think this is a relative statement... relative to the language constraints...

attribute/yoti.go
LKDraytonBrown: "it made testing a lot more difficult - having to create these protobuf attributes instead of a new type." - I don't see how this is different, you will still need to construct the types for testing... also you could build test helpers. Core has some Go code for doing this, maybe we could share?

LKDraytonBrown: can make the map right here: 
```
func GetMIMEType(attributeType AttrType) string {
    return map[AttrType]string{
                {AttrTypeTime: "text/plain; charset=UTF-8"},
               ...
    }

File: attribute/generic.go (hunk 25 above)
Comment (echarrod) [reply attribute/generic.go]: Yes, it's a question of how far you take "as much as possible". @bucky-boy - we're sticking with string representations, right?

File: attribute/yoti.go (hunk 26 above)
Comment (echarrod) [reply attribute/yoti.go]: > "it made testing a lot more difficult - having to create these protobuf attributes instead of a new type." - I don't see how this is different, you will still need to construct the types for testing... also you could build test helpers. Core has some Go code for doing this, maybe we could share?
> With regards to the second part, I'm not sure why including the 'raw' anchor is an issue, the only interesting information in there is what we are going to share and the YOTI signatures to verify that YOTI has processed the data.... this is all OK for the public to see.

If you have some Go code for building test helpers, that would be much appreciated. If you could reference them in [the ticket for looking at using Protobuf types]( https://lampkicking.atlassian.net/browse/SDK-624), then we can get this PR merged, and look at doing this as part of that ticket.

Comment (echarrod) [reply attribute/yoti.go]: > can make the map right here:
> 
> ```
> func GetMIMEType(attributeType AttrType) string {
>     return map[AttrType]string{
>                 {AttrTypeTime: "text/plain; charset=UTF-8"},
>                ...
>     }
> Also notice the more specific type being passed in
> ```

I did have the more specific type originally, but gometalinter suggested I change it  to `fmt.Stringer`(not sure why). I'll change it back

Comment (echarrod) [reply attribute/yoti.go]: image

Comment (LKDraytonBrown) [reply attribute/yoti.go]: oh, ok. That's interesting. Not sure why its making that suggestion. Going to ask @simon-yoti

Comment (LKDraytonBrown) [reply attribute/yoti.go]: Is there any more to this warning message?

Comment (echarrod) [reply attribute/yoti.go]: No, that is it unfortunately

Comment (simon-yoti) [reply attribute/yoti.go]: I think this is because of the Go mantra
>return types, accept interfaces.

So the linter looks to see if an argument implements an interface, and if it does suggests you use that instead. In this case it's wrong; `AttrType` does have a `String() string` method and so implements `fmt.Stringer`, but the behaviour of the function has nothing to with what interface it implements (it just happens to use the `String` method in an error message). Linters aren't always right, that's why they're linters.

Comment (echarrod) [reply attribute/yoti.go]: Thanks, ok I'll leave it as using `AttrType`, not `fmt.Stringer`

File: conversion.go (hunk 21 above)
Comment (echarrod) [reply conversion.go]: I've addressed this in a commit to this branch now - https://github.com/getyoti/yoti-go-sdk/pull/22/commits/f412515397e65b30a07873d872dc0c09415cb4f0

File: attribute/timeattribute.go (hunk 18 above)
Comment (echarrod) [reply attribute/timeattribute.go]: Have returned `nil, err` in the case where there has been an error parsing now

File: yotiprofile.go (hunk 27 above)
Comment (echarrod) [reply yotiprofile.go]: We have had to redefine them as we want anyone to be able to check out the SDK and build it, so we can't include private APIs

File: image.go [hunk 29]
```diff
@@ -1,24 +1,29 @@
...
+//Deprecated: Will be removed in v3.0.0 - use attribute.ContentType instead. ImageType Image format
 type ImageType int
 
 const (
 	//ImageTypeJpeg JPEG format
 	ImageTypeJpeg ImageType = 1 + iota
 	//ImageTypePng PNG format
 	ImageTypePng
+	//ImageTypeOther Other image formats
+	ImageTypeOther
```
bucky-boy: I don't think we support any other image formats.
If and when we do, they'll have names.

echarrod: This is there just to have an option for the imageType if it doesn't match either PNG or JPEG. We don't ever want this to be used, since we will introduce a new type, it's there just to cover the "just in case" scenario

PR #23, opened 2018-11-09
[SDK-443] Add Sources & Verifiers

echarrod:
- Add sources & verifiers
- Tidying up example project
- Update browser.js

I found a couple of issues with the method descriptions, so I will review all of these before releasing.


README.md
```diff
@@ -135,6 +135,24 @@ userProfile.GetAttribute("age_over:18").Value.(string)
...
+givenNamesAnchors := userProfile.GivenNames().Anchors
+givenNamesSources := userProfile.GivenNames().Sources
+givenNamesVerifiers := userProfile.GivenNames().Verifiers
+```
+You can also retrieve further properties from these respective anchors in the following way:
+```Go
+givenNamesFirstAnchor := userProfile.GivenNames().Anchors[0]
+
+anchorType := givenNamesFirstAnchor.Type
+originServerCerts := givenNamesFirstAnchor.OriginServerCerts()
```
bucky-boy: 1. No point in directing them to _originServerCerts_, because at the moment they can't do anything with that info.
2. We should offer some kind of explanation for what the _type_, _subType_ & _value_ are.  See the Java README

attribute/genericattribute.go
```diff
@@ -11,8 +11,10 @@ import (
 //GenericAttribute is a Yoti attribute which returns a generic value
 type GenericAttribute struct {
 	*yotiprotoattr_v3.Attribute
-	Value   interface{}
-	Anchors []*anchor.Anchor
+	Value     interface{}
+	Anchors   []*anchor.Anchor
```
bucky-boy: Are these lists immutable?

attribute/image.go
```diff
@@ -30,19 +30,22 @@ var mimeTypeMap = map[ImageType]string{
...
-	log.Printf("Unable to find a matching MIME type for value type %q", imageType)
-	return
+	return "", fmt.Errorf("Unable to find a matching MIME type for value type %q", imageType)
 }
 
 // Base64URL is the Image encoded as a base64 URL
 func (image *Image) Base64URL() (string, error) {
 	base64EncodedImage := base64.StdEncoding.EncodeToString(image.Data)
-	contentType := GetMIMEType(image.Type)
+	contentType, err := GetMIMEType(image.Type)
```
bucky-boy: Defining our own enum for MIME type doesn't add anything.  We can just keep the type on the Image as a String.

README.md
echarrod: So best to remove the accessor OriginServerCerts() for now too? And we can always add it in later?

attribute/genericattribute.go
echarrod: Doesn't look like it's currently possible to add readonly modifiers from these proposals [[1]](https://github.com/golang/go/issues/20443) [[2]](https://github.com/golang/go/issues/22876)

We could remove them as fields on the struct, and instead expose GetAnchors, GetSources and GetVerifier methods. I did consider this, but it would mean quite a bit of duplication as we would need to define this on each of the different attribute types in the current way it's set up.

attribute/image.go
echarrod: Will change to string

README.md
bucky-boy: No, just drop it from the docs.  No harm in leaving access to the certs in place - but there's no point drawing attention to them when they can't be used for anything (yet)

attribute/genericattribute.go
bucky-boy: This is a shame.
But we do to expose immutable objects from the SDKs, somehow...

echarrod:
In the last commit I changed the anchors, sources & verifiers fields to be unexported so that we can restrict access to them through accessor methods. This has the disadvantage of a slight bit of duplication in each of the attribute classes, but Mike and I had a chat and believe this is better than having the potential for unwanted modifications of these three slices.

anchor/anchors.go
```diff
@@ -76,3 +76,22 @@ func (a Anchor) SubType() string {
...
 }
+
+// GetSources returns the anchors which identify how and when an attribute value was acquired.
+func GetSources(anchors []*Anchor) (sources []*Anchor) {
+	return filterAnchors(anchors, AnchorTypeSource)
+}
+
+// GetVerifiers returns the anchors which identify how and when an attribute value was verified by another provider.
+func GetVerifiers(anchors []*Anchor) (sources []*Anchor) {
+	return filterAnchors(anchors, AnchorTypeVerifier)
```
bucky-boy: Since we already have _Type_ as a string on each Anchor struct (as taken from the certificate), can't we just filter on that?  What's the additional _type_ enum for?

attribute/genericattribute.go
```diff
@@ -49,12 +51,33 @@ func NewGeneric(a *yotiprotoattr_v3.Attribute) *GenericAttribute {
...
+	parsedAnchors := anchor.ParseAnchors(a.Anchors)
+
 	return &GenericAttribute{
 		Attribute: &yotiprotoattr_v3.Attribute{
 			Name:        a.Name,
 			ContentType: a.ContentType,
 		},
-		Value:   value,
-		Anchors: anchor.ParseAnchors(a.Anchors),
+		Value:     value,
```
bucky-boy: I'm not clear how this works in Go, but we need to sure that _Value_ is immutable too.  We don't really want people to have any ability to overwrite the values on the attributes we return.

anchor/anchors.go
bucky-boy: *Update*, I'm a bit wrong here, we've just talked this through....

The _anchor type_ comes to us as an 'extensionOId' e.g. '1.3.6.1.4.1.47127.1.1.1'.  We are not given a nice, neat String.  The String's are introduced by the SDK to make the OId more meaningful.

The apprach here matches the Java SDK - map the OId to an Enum as a convenient way to convert that OId into a nice string.  If we ever get an OId we don't know about, we map it to _unknown_.

We might to want to improve on this so I've created SDK-683, but for now the work done here is good to go.

README.md
```diff
@@ -130,11 +132,45 @@ if err != nil {
...
+givenNamesVerifiers := userProfile.GivenNames().Verifiers()
+```
+
+You can also retrieve further properties from these respective anchors in the following way:
+
+```Go
+givenNamesFirstAnchor := userProfile.GivenNames().Anchors()[0]
+
+anchorType := givenNamesFirstAnchor.Type
+signedTimestamp := givenNamesFirstAnchor.SignedTimestamp().Timestamp
```
sidibos: Is Timestamp a DateTime Object?
If so, it will be good to add it as a comments.

Like this one -> // object, the actual timestamp as a *time.Time can be called with

anchor/anchors.go
```diff
@@ -76,3 +76,22 @@ func (a Anchor) SubType() string {
...
+}
+
+func filterAnchors(anchors []*Anchor, anchorType Type) (result []*Anchor) {
+	for _, v := range anchors {
+		if v.Type == anchorType {
```
sidibos: Does the Anchor Object include the AnchorType?
If so what's the value of the type e.g 'source' or the old e.g '1.3.6.1.4.1.47127.1.1.1'?

attribute/genericattribute.go
```diff
@@ -11,8 +11,10 @@ import (
 //GenericAttribute is a Yoti attribute which returns a generic value
 type GenericAttribute struct {
 	*yotiprotoattr_v3.Attribute
-	Value   interface{}
-	Anchors []*anchor.Anchor
+	value     interface{}
+	anchors   []*anchor.Anchor
```
sidibos: Does this hold a list of Anchors e.g [AnchorObj1, AnchorObj2, ...] or a Map of Anchors as ['1.3.6.1.4.1.47127.1.1.1' => AnchorObj, ...]?

README.md
echarrod: Yes it is a *time.Time. Instead of a comment, I'm going to specify the type explicitly on the LHS. Even though the compiler gives you a warning that the type can be inferred, and it's unnecessary, I think you're right in it's worthwhile to show the users what type to expect when we're displaying these in the README.

attribute/genericattribute.go
echarrod: Yes this is a slice, a bit like a list - there's basically nothing a list can do that a slice can't (in Go), but lists have the disadvantage of not being strongly typed. You can have a [map](https://blog.golang.org/go-maps-in-action), but I don't think that is necessary here

anchor/anchors.go
echarrod: Discussed in https://lampkicking.atlassian.net/browse/SDK-683: having enum type is fine for the foreseeable future, and we won't expose OID

PR #24, opened 2018-11-16
[SDK-513]: Remove version from protobuf packages

echarrod:
- The v3 package names weren't released, that change is currently in `development` branch
- Did a few of the quick wins for `gometalinter` issues, have left the non-trivial ones for the next release


yotiprotoattr/List.proto
```diff
@@ -1,13 +1,13 @@
 syntax = "proto3";
 
-package Yoti.Protobuf.attrpubapi_v3;
+package attrpubapi_v1;
```
sidibos: Should this be `attrpubapi` instead if we are getting rid of the version?

yotiprotocom/EncryptedData.proto
```diff
@@ -1,11 +1,11 @@
 syntax = "proto3";
 
-package Yoti.Protobuf.compubapi_v3;
+package compubapi_v1;
```
sidibos: Same here?

attribute/jsonattribute.go
```diff
@@ -5,20 +5,20 @@ import (
...
 type JSONAttribute struct {
-	*yotiprotoattr_v3.Attribute // Value returns the value of a JSON attribute in the form of an interface
-	value                       interface{}
-	anchors                     []*anchor.Anchor
-	sources                     []*anchor.Anchor
-	verifiers                   []*anchor.Anchor
+	*yotiprotoattr.Attribute // Value returns the value of a JSON attribute in the form of an interface
+	value                    interface{}
+	anchors                  []*anchor.Anchor
+	sources                  []*anchor.Anchor
```
sidibos: Do you need `sources` and `verifiers` on top of `anchors`?

yotiprotoattr/Attribute.proto
```diff
@@ -1,11 +1,11 @@
 syntax = "proto3";
 
-package Yoti.Protobuf.attrpubapi_v3;
+package attrpubapi_v1;
 
 option java_package = "com.yoti.attrpubapi_v3";
```
qzagarese: Should you remove it here too?

yotiprotoattr/Signing.proto
```diff
@@ -1,13 +1,13 @@
 syntax = "proto3";
 
-package Yoti.Protobuf.attrpubapi_v3;
+package attrpubapi_v1;
 
 import "Attribute.proto";
 
 option java_package = "com.yoti.attrpubapi_v3";
```
qzagarese: same here

yotiprotocom/EncryptedData.proto
```diff
@@ -1,11 +1,11 @@
...
 
 option java_package = "com.yoti.compubapi_v3";
```
qzagarese: same here

yotiprotocom/SignedTimestamp.proto
```diff
@@ -1,11 +1,11 @@
 ﻿syntax = "proto3";
 
-package Yoti.Protobuf.compubapi_v3;
+package compubapi_v1;
 
 option java_package = "com.yoti.compubapi_v3";
```
qzagarese: same here

yotiprotoattr/List.proto
echarrod: I'm just returning it to what it was before, and what is currently present on https://github.com/lampkicking/sdk-protobuf. 

I believe this line isn't actually used anyway, since we are overriding it with the specific go package name below

yotiprotoattr/Attribute.proto
echarrod: This won't ever be used by Java, since we're going to refer to the main lampkicking repo's proto files, but I guess there's no harm in changing it, I'll update it.

attribute/jsonattribute.go
sidibos: I meant only pass in anchors and parse sources and verifiers from that.

echarrod: Ah yes good point, will change

PR #25, opened 2018-11-20
[SDK-609]: Create Yoti v2 module

echarrod:
I've gone with the "Major branch" strategy outlined in the [Go wiki on Modules ](https://github.com/golang/go/wiki/Modules#releasing-modules-v2-or-higher). This allows us to carry on working on master, and if we do need to change anything for version 2 (after we've released version 3), then we can make a branch for v2, and make any bug fixes there. 
I'll add further detail to a wiki shortly, including what we'll need to do with each minor and major release going forwards.

When we release the major version, it will just be a case of renaming the module, and changing the v2s to v3s (or using [this handy tool](https://github.com/marwan-at-work/mod)). 

Also updated some fields to make them read only. 

This will require an update to the developer docs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,9 @@ Yoti also allows you to enable user details verification from your mobile app by
 
 ## Installing the SDK
 
-To download and install the Yoti SDK and its dependencies, simply run the following command from your terminal:
+_As of version **2.3.0**, [modules](https://github.com/golang/go/wiki/Modules) are used. This means it's not neccesary to get a copy or fetch all dependencies as instructed below, as the Go toolchain will fetch them as neccesary. You can simply add a `require github.com/getyoti/yoti-go-sdk/v2` to go.mod._
+
+To download and install the Yoti SDK and its dependencies, run the following command from your terminal:
 
 ```Go
 go get "github.com/getyoti/yoti-go-sdk/v2"
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ Yoti also allows you to enable user details verification from your mobile app by
 To download and install the Yoti SDK and its dependencies, simply run the following command from your terminal:
 
 ```Go
-go get "github.com/getyoti/yoti-go-sdk"
+go get "github.com/getyoti/yoti-go-sdk/v2"
 ```
 
 ## SDK Project import
@@ -64,7 +64,7 @@ go get "github.com/getyoti/yoti-go-sdk"
 You can reference the project URL by adding the following import:
 
 ```Go
-import "github.com/getyoti/yoti-go-sdk"
+import "github.com/getyoti/yoti-go-sdk/v2"
 ```
 
 ## Configuration
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ Yoti also allows you to enable user details verification from your mobile app by
 
 ## Installing the SDK
 
-_As of version **2.3.0**, [modules](https://github.com/golang/go/wiki/Modules) are used. This means it's not neccesary to get a copy or fetch all dependencies as instructed below, as the Go toolchain will fetch them as neccesary. You can simply add a `require github.com/getyoti/yoti-go-sdk/v2` to go.mod._
+_As of version **2.4.0**, [modules](https://github.com/golang/go/wiki/Modules) are used. This means it's not necessary to get a copy or fetch all dependencies as instructed below, as the Go toolchain will fetch them as necessary. You can simply add a `require github.com/getyoti/yoti-go-sdk/v2` to go.mod._
 
 To download and install the Yoti SDK and its dependencies, run the following command from your terminal:
 
--- a/anchor/anchorparser.go
+++ b/anchor/anchorparser.go
@@ -8,8 +8,8 @@ import (
 	"fmt"
 	"log"
 
-	"github.com/getyoti/yoti-go-sdk/yotiprotoattr"
-	"github.com/getyoti/yoti-go-sdk/yotiprotocom"
+	"github.com/getyoti/yoti-go-sdk/v2/yotiprotoattr"
+	"github.com/getyoti/yoti-go-sdk/v2/yotiprotocom"
 	"github.com/golang/protobuf/proto"
 )
 
--- a/attribute/stringattribute.go
+++ b/attribute/stringattribute.go
@@ -1,8 +1,8 @@
 package attribute
 
 import (
-	"github.com/getyoti/yoti-go-sdk/anchor"
-	"github.com/getyoti/yoti-go-sdk/yotiprotoattr"
+	"github.com/getyoti/yoti-go-sdk/v2/anchor"
+	"github.com/getyoti/yoti-go-sdk/v2/yotiprotoattr"
 )
 
 // StringAttribute is a Yoti attribute which returns a string as its value
--- a/examples/go.mod
+++ b/examples/go.mod
@@ -1,0 +1,3 @@
+module github.com/getyoti/yoti-go-sdk/examples
+
+require github.com/joho/godotenv v1.3.0
--- a/yoti_test.go
+++ b/yoti_test.go
@@ -12,9 +12,9 @@ import (
 	"testing"
 	"time"
 
-	"github.com/getyoti/yoti-go-sdk/anchor"
-	"github.com/getyoti/yoti-go-sdk/attribute"
-	"github.com/getyoti/yoti-go-sdk/yotiprotoattr"
+	"github.com/getyoti/yoti-go-sdk/v2/anchor"
+	"github.com/getyoti/yoti-go-sdk/v2/attribute"
+	"github.com/getyoti/yoti-go-sdk/v2/yotiprotoattr"
 	"github.com/golang/protobuf/proto"
 	"github.com/google/go-cmp/cmp"
 )
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,14 +1,16 @@
 # Contributing
 
-The command `go get "github.com/getyoti/yoti-go-sdk"` downloads the Yoti package, along with its dependencies, and installs it.
+The command `go get "github.com/getyoti/yoti-go-sdk/v2"` downloads the Yoti package, along with its dependencies, and installs it.
 
 ## Commit Process
 
+1) `go build` builds the package (then discards the results)
 1) `goimports` formats the code and sanitises imports
 1) `go vet` reports suspicious constructs
 1) `go test` to run the tests
 1) `go test -race` detects for race conditions
 1) `golangci-lint run` for [GolangCI-Lint](https://github.com/golangci/golangci-lint)
+1) `go mod tidy` prunes any no-longer-needed dependencies from `go.mod`, and adds any dependencies needed
 
 ## VS Code
 
--- a/anchor/anchors.go
+++ b/anchor/anchors.go
@@ -3,7 +3,7 @@ package anchor
 import (
 	"crypto/x509"
 
-	"github.com/getyoti/yoti-go-sdk/yotiprotocom"
+	"github.com/getyoti/yoti-go-sdk/v2/yotiprotocom"
 )
 
 // Anchor is the metadata associated with an attribute. It describes how an attribute has been provided
--- a/anchor/signedtimestamp.go
+++ b/anchor/signedtimestamp.go
@@ -3,7 +3,7 @@ package anchor
 import (
 	"time"
 
-	"github.com/getyoti/yoti-go-sdk/yotiprotocom"
+	"github.com/getyoti/yoti-go-sdk/v2/yotiprotocom"
 )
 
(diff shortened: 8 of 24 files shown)


yoti_test.go
```diff
@@ -1018,24 +1018,6 @@ func TestAnchorParser_YotiAdmin(t *testing.T) {
 	AssertServerCertSerialNo(t, "256616937783084706710155170893983549581", actualSerialNo)
 }
 
-func TestAnchors_Unknown(t *testing.T) {
```
echarrod: Since anchor type is unexported now, we can't create this unknown anchor in the way that we're doing in this test. I spoke with Mike and have created https://lampkicking.atlassian.net/browse/SDK-686 to reimplement this test once we have a mocking framework. Drayton also mentioned Core had some testing tools, which might solve this.

echarrod:
Accompanying decision log: https://lampkicking.atlassian.net/wiki/spaces/SDK/pages/917831708/Version+Management+for+Go+SDK

pn-santos:
Regarding the modules part, LGTM :+1:

yoti_test.go
```diff
@@ -904,12 +904,12 @@ func TestAnchorParser_Passport(t *testing.T) {
 		t.Error("Anchors and Sources should be the same when there is only one Source")
 	}
 
-	if actualAnchor.Type != anchor.AnchorTypeSource {
-		t.Errorf("Parsed anchor type is incorrect. Expected: %q, actual: %q", anchor.AnchorTypeSource, actualAnchor.Type)
+	if actualAnchor.Type() != anchor.AnchorTypeSource {
+		t.Errorf("Parsed anchor type is incorrect. Expected: %q, actual: %q", anchor.AnchorTypeSource, actualAnchor.Type())
```
sjones303: Consider splitting long lines, currently have to scroll sideways on GitHub to view the whole line in quite a few places.

README.md
```diff
@@ -56,15 +56,15 @@ Yoti also allows you to enable user details verification from your mobile app by
 To download and install the Yoti SDK and its dependencies, simply run the following command from your terminal:
 
 ```Go
-go get "github.com/getyoti/yoti-go-sdk"
+go get "github.com/getyoti/yoti-go-sdk/v2"
```
sjones303: Consider mentioning modules are now used, so it's not necessary get a copy or fetch all dependencies as the Go toolchain will fetch them as necessary. Unless someone wishes to run the examples or make a contribution it's sufficient to just start using it/add a `require` to `go.mod` and the toolchain will take care of the rest.

activitydetails.go
```diff
@@ -0,0 +1,14 @@
...
+
+//ActivityDetails represents the result of an activity between a user and the application
+type ActivityDetails struct {
+	UserProfile  Profile
+	rememberMeID string
+}
+
+// RememberMeID is a unique identifier Yoti assigns to your user, but only for your app
+// if the same user logs into your app again, you get the same id
+// if she/he logs into another application, Yoti will assign a different id for that app
```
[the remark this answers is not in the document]
echarrod: Thanks, makes sense. I'll change to either have a blank line, or make sure they are separated with a full stop.

anchor/anchorparser.go
```diff
@@ -28,18 +28,20 @@ func ParseAnchors(protoAnchors []*yotiprotoattr.Anchor) []*Anchor {
 
 		for _, cert := range parsedCerts {
 			for _, ext := range cert.Extensions {
-				var ae anchorExtension
 				if sourceOID.Equal(ext.Id) {
 					anchorType = AnchorTypeSource
+
+					extension := unmarshalExtension(ext.Value)
+					if extension != "" {
+						extensions = append(extensions, extension)
```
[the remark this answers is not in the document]
sjones303: That is in essence what I'm saying.

It's a shame the documentation from the protobufs has been stripped, the documentation I've seen clearly states the first certificate in the chain will have the source/verifier extension, this means if all we're interested in is the source/verifier (i.e. not verifying the chain) then we only need to process/parse the first certificate.

In addition to that a certificate may have many extensions (e.g. key use), but the leaf certificate will only have one Yoti source or one Yoti verifier extension, it may have other non-Yoti extensions. Therefore to find the anchor type and the source/verifier value we just need to loop through the extensions until a Yoti OID is found.

This code currently works because there is only ever one source or one verifier extension in the chain. If it were to ever process a malformed anchor (i.e. multiple source/verifier extensions) then the values slice will have multiple entries, but the type will only be of the last extension processed. This isn't too much of an issue if the anchor was trying to express multiple verifications (type would be verification, with multiple values), but if there was a mixture then both the type and values would be meaningless.

examples/go.mod
```diff
@@ -0,0 +1,3 @@
+module github.com/getyoti/yoti-go-sdk/examples
+
+require github.com/joho/godotenv v1.3.0
```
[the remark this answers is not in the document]
echarrod: So it it best to leave this as is, and then we add it as a requirement once v2.3.0 is tagged?

anchor/anchorparser.go
```diff
@@ -2,45 +2,47 @@ package anchor
...
-					extensions = append(extensions, ae.Extension)
-				} else {
-					log.Printf("Error unmarshalling anchor extension: %q", err)
+				var (
+					value string
+					err   error
+				)
+				anchorType, value, err = parseExtension(ext)
+				if err != nil {
+					log.Printf("error parsing anchor extension, %v", err)
```
sjones303: I know I suggested this, but I've realised we will now log an error for other extensions, which is incorrect. Maybe it would be better to change the `default` case in `parseExtension` to be:
```go
default:
	return anchorType, "", nil
```
Then you can do:
```go
if err != nil {
	⋮
} else if anchorType == AnchorTypeUnknown {
	continue
}
⋮
```

anchor/anchorparser.go
```diff
@@ -52,6 +54,45 @@ func ParseAnchors(protoAnchors []*yotiprotoattr.Anchor) []*Anchor {
...
+		anchorType = AnchorTypeSource
+	case ext.Id.Equal(verifierOID):
+		anchorType = AnchorTypeVerifier
+	default:
+		return anchorType, "", errors.New("unknown anchor type")
+	}
+
+	var ae anchorExtension
+	_, err = asn1.Unmarshal(ext.Value, &ae)
+
```
sjones303: It generally make things easier to follow if there's no space between the function call and the error handling.

yoti_test.go
```diff
@@ -217,7 +215,8 @@ func TestYotiClient_ParseWithoutProfile_Success(t *testing.T) {
 			result = &httpResponse{
 				Success:    true,
 				StatusCode: 200,
-				Content:    `{"receipt":{"wrapped_receipt_key": "` + wrappedReceiptKey + `",` + otherPartyProfileContent + `"remember_me_id":"` + rememberMeID + `", "sharing_outcome":"SUCCESS"}}`}
+				Content: `{"receipt":{"wrapped_receipt_key": "` + wrappedReceiptKey + `",` +
```
sjones303: Has this been run through `goimports`? It looks like the alignment is off.

yoti_test.go
```diff
@@ -12,16 +12,17 @@ import (
...
 const sdkID = "fake-sdk-id"
+const wrappedReceiptKey = "kyHPjq2+Y48cx+9yS/XzmW09jVUylSdhbP+3Q9Tc9p6bCEnyfa8vj38AIu744RzzE+Dc4qkSF21VfzQKtJVILfOXu5xRc7MYa5k3zWhjiesg/gsrv7J4wDyyBpHIJB8TWXnubYMbSYQJjlsfwyxE9kGe0YI08pRo2Tiht0bfR5Z/YrhAk4UBvjp84D+oyug/1mtGhKphA4vgPhQ9/y2wcInYxju7Q6yzOsXGaRUXR38Tn2YmY9OBgjxiTnhoYJFP1X9YJkHeWMW0vxF1RHxgIVrpf7oRzdY1nq28qzRg5+wC7cjRpS2i/CKUAo0oVG4pbpXsaFhaTewStVC7UFtA77JHb3EnF4HcSWMnK5FM7GGkL9MMXQenh11NZHKPWXpux0nLZ6/vwffXZfsiyTIcFL/NajGN8C/hnNBljoQ+B3fzWbjcq5ueUOPwARZ1y38W83UwMynzkud/iEdHLaZIu4qUCRkfSxJg7Dc+O9/BdiffkOn2GyFmNjVeq754DCUypxzMkjYxokedN84nK13OU4afVyC7t5DDxAK/MqAc69NCBRLqMi5f8BMeOZfMcSWPGC9a2Qu8VgG125TuZT4+wIykUhGyj3Bb2/fdPsxwuKFR+E0uqs0ZKvcv1tkNRRtKYBqTacgGK9Yoehg12cyLrITLdjU1fmIDn4/vrhztN5w="
```
sjones303: Consider placing these in a single `const (…)` block, in my opinion it is cleaner, and can allow the whole block to be documented if necessary.

examples/go.mod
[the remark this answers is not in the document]
sjones303: Yes, that's right.

yoti_test.go
echarrod: I currently go through the steps on the [CONTRIBUTING.md](https://github.com/getyoti/yoti-go-sdk/blob/master/CONTRIBUTING.md) page, as well as doing `gometalinter`, so we're doing `go fmt`. Would you suggest replacing `go fmt` with `goimports`? 

In any case, this shouldn't have been missed by `go fmt`, doesn't seem to be changed when I run it now either. I'll investigate

anchor/anchorparser.go
echarrod: Yes that's sensible, will change

anchor/anchorparser.go
echarrod: Sure. Is it just your preference on using switch statements here instead of:
```
if err != nil {
    return anchorType, "", fmt.Errorf("unable to unmarshal extension: %v", err)
} else if len(ae.Extension) == 0 {
    return anchorType, "", errors.New("empty extension")
}

val = ae.Extension
```

I've always used the above when we're not switching on a definitive object type

yoti_test.go
echarrod: Yep I agree, that would be cleaner

yoti_test.go
sjones303: Yeah, I had a play too. Since `goimports` essentially wraps `go fmt` it has the same behaviour, wasn't aware of this edge case.

It would be worth some time in the future looking into using `goimports`, it does all that `go fmt` does, but also sanitises imports (e.g. removes unused ones to prevent build failures down the line).

anchor/anchorparser.go
sjones303: They're both pretty much the same (I've not had a dig to see if they are optimised differently). It's a fairly common pattern to use a `switch` for more complex error (or just return values) processing, in my opinion since the logic can be expressed as scenarios (i.e. cases) it's cleaner to write it that way, rather than chaining `if`s. Unless the processing code is hard to follow as written I won't tend to suggest changing between `if`/`switch`.

echarrod:
I have reverted the addition of modules for this minor release. We will look at adding them in the next release.

attribute/stringattribute.go
```diff
@@ -5,14 +5,14 @@ import (
...
 type StringAttribute struct {
 	*yotiprotoattr.Attribute
 	value   string
 	anchors []*anchor.Anchor
 }
 
-//NewString creates a new String attribute
+// NewString creates a new String attribute
 func NewString(a *yotiprotoattr.Attribute) *StringAttribute {
 	parsedAnchors := anchor.ParseAnchors(a.Anchors)
```
sidibos: Not sure about this approach, maybe the parser shoddy have been in a different module. Like AnchorParser or Converter.

anchor/anchorparser.go
```diff
@@ -14,33 +17,34 @@ type anchorExtension struct {
...
-					log.Printf("Error unmarshalling anchor extension: %q", err)
+				var (
+					value string
+					err   error
+				)
+				anchorType, value, err = parseExtension(ext)
+				if err != nil {
+					log.Printf("error parsing anchor extension, %v", err)
+					continue
+				} else if anchorType == AnchorTypeUnknown {
```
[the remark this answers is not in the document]
echarrod: We will be parsing the unknown anchors, but we won't be parsing the extensions on those anchors. See the discussion on SDK-683

attribute/stringattribute.go
echarrod: the `anchor` at the start is a package qualifier - i.e. we are accessing the `ParseAnchors` function in the `anchor` package.

sidibos: 👍

anchor/anchorparser.go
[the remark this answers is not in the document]
sidibos: 👍

PR #31, opened 2019-03-05
SDK-761: Document Images Attribute

echarrod:
- Add new DocumentImages attribute, which is of type `ImageSliceAttribute`
- Bit of tidying up as I went, might be a better idea to look at the individual commits
- Test has a raw attribute which contains two document images


attribute/genericattribute.go
```diff
@@ -17,48 +17,52 @@ type GenericAttribute struct {
...
 	case yotiprotoattr.ContentType_STRING:
-		value = string(a.Value)
+		result = string(byteValue)
 
 	case yotiprotoattr.ContentType_JPEG,
 		yotiprotoattr.ContentType_PNG,
+		yotiprotoattr.ContentType_MULTI_VALUE,
 		yotiprotoattr.ContentType_UNDEFINED:
-		value = a.Value
+		result = byteValue
```
qzagarese: I'm not convinced by this.
Here you need a separate case that triggers the recursive call.
If it's an image, you correctly return the binary, but if it's a MULTI_VALUE, you need to do the following:

1. Unmarshal the binary a MutliValue protobuf https://github.com/lampkicking/core-api/blob/master/attrpubapi_v1/Attribute.proto#L39
2. Create a slice that is as long as the number of `values` in the unmarshalled proto.
3. Iterate through the values and invoke `parseValue` passing the `contentType` and `data`  from the i-th value. 
4. Assign the result of the call to the corresponding element in the slice.

echarrod:
- Added general MultiValueAttribute class which, along with the other standard attribute methods, returns:
  - items, which are the multiValues (calling them values would be too many `values`). These items have interfaces (because of a lack of generics), which must be cast to the desired type based on the `ContentType` property
- The above class will only be used when we use profile.GetAttribute(attributeName)
- I made ParseMultiValue exported, as it is needed for when we retrieve images with the above method
- I've created SDK-849 to change how these generic attributes are parsed, to be completed with v3.0.0
- The best way to understand how the multivalues work is looking at the tests. The nested multivalue is a bit convoluted, but I didn't see another way to test it as thoroughly (we also have SDK-537 to split up the tests into separate files based on the packages they test)
- CreateImageSlice effectively has a built-in filter (in a similar fashion to Node and Java), by not adding non-Image values to the slice
- I tried to parse the multivalues non-recursively, but it ended up looking pretty convoluted and messy, so I've kept the recursive method in for now

attribute/multivalueattribute.go
```diff
@@ -0,0 +1,88 @@
...
+	protoMultiValueStruct := unmarshallMultiValue(data)
+
+	for _, multiValueItem := range protoMultiValueStruct.Values {
+		if multiValueItem.ContentType == yotiprotoattr.ContentType_MULTI_VALUE {
+			var parsedInnerMultiValueItems []*Item = ParseMultiValue(multiValueItem.Data)
+			var parsedInnerMultiValueItemsSingleItem = &Item{
+				contentType: yotiprotoattr.ContentType_MULTI_VALUE,
+				value:       parsedInnerMultiValueItems,
+			}
+			attributeItems = append(attributeItems, parsedInnerMultiValueItemsSingleItem)
```
qzagarese: If both `parsedInnerMultiValueItemsSingleItem` and `value` are of type `Item`, you can bring this line and line 50 outside the if else.

PR #34, opened 2019-03-18
SDK-756: Add further tests for Multi Value parsing

echarrod:
- when there's an error parsing any of the multi value contents, don't return the partial attribute

### DIFF
--- a/attribute/multivalueattribute.go
+++ b/attribute/multivalueattribute.go
@@ -1,7 +1,7 @@
 package attribute
 
 import (
-	"log"
+	"fmt"
 
 	"github.com/getyoti/yoti-go-sdk/v2/anchor"
 	"github.com/getyoti/yoti-go-sdk/v2/yotiprotoattr"
@@ -17,7 +17,11 @@ type MultiValueAttribute struct {
 
 // NewMultiValue creates a new MultiValue attribute
 func NewMultiValue(a *yotiprotoattr.Attribute) (*MultiValueAttribute, error) {
-	var attributeItems []*Item = ParseMultiValue(a.Value)
+	attributeItems, err := parseMultiValue(a.Value)
+
+	if err != nil {
+		return nil, err
+	}
 
 	return &MultiValueAttribute{
 		Attribute: &yotiprotoattr.Attribute{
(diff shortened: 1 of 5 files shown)


attribute/multivalueattribute.go
```diff
@@ -29,40 +33,55 @@ func NewMultiValue(a *yotiprotoattr.Attribute) (*MultiValueAttribute, error) {
 	}, nil
 }
 
-// ParseMultiValue recursively unmarshals and converts Multi Value bytes into a slice of Items
-func ParseMultiValue(data []byte) []*Item {
+// parseMultiValue recursively unmarshals and converts Multi Value bytes into a slice of Items
+func parseMultiValue(data []byte) ([]*Item, error) {
```
echarrod: (This hasn't been released yet, so changing it from exported to unexported is not a breaking change)

yoti_test.go
```diff
@@ -1254,7 +1291,29 @@ func TestNestedMultiValue(t *testing.T) {
 		Anchors:     []*yotiprotoattr.Anchor{},
 	}
 
-	multiValueAttribute, err := attribute.NewMultiValue(protoAttribute)
+	profile := createProfileWithSingleAttribute(protoAttribute)
+
+	if profile.GetAttribute(attributeName) != nil {
+		t.Error("Excpected to retrieve null attribute when there is erroneous data within the attribute value")
```
davidgrayston:
```suggestion
		t.Error("Expected to retrieve null attribute when there is erroneous data within the attribute value")
```

attribute/parser.go
```diff
@@ -1,45 +1,44 @@
...
 	"github.com/getyoti/yoti-go-sdk/v2/yotiprotoattr"
 )
 
-func parseValue(contentType yotiprotoattr.ContentType, byteValue []byte) (result interface{}) {
+func parseValue(contentType yotiprotoattr.ContentType, byteValue []byte) (interface{}, error) {
 	switch contentType {
 	case yotiprotoattr.ContentType_DATE:
 		parsedTime, err := time.Parse("2006-01-02", string(byteValue))
+
 		if err == nil {
```
davidgrayston: Should we return in here? Or change to `!== nil` and return the error?

echarrod: Good spot, not sure why it's like this!

yoti_test.go
```diff
@@ -628,6 +628,28 @@ func TestProfile_GetAttribute_String(t *testing.T) {
...
+		Name:        attrConstGender,
+		Value:       attributeValue,
+		ContentType: yotiprotoattr.ContentType_STRING,
+		Anchors:     []*yotiprotoattr.Anchor{},
+	}
+
+	profile := createProfileWithSingleAttribute(attr)
+	att := profile.Gender()
+
+	if !cmp.Equal(att.Value(), attributeValueString) {
```
davidgrayston: Could we make use of https://godoc.org/gotest.tools/assert? Perhaps for another ticket/PR

echarrod: I've added SDK-865 for that 👍

yoti_test.go
```diff
@@ -1226,20 +1286,19 @@ func TestNewMultiValue(t *testing.T) {
 	assertIsExpectedDocumentImagesAttribute(t, documentImagesAttributeItems, multiValueAttribute.Anchors()[0])
 }
 
-func TestNestedMultiValue(t *testing.T) {
-	var innerMultiValueProtoValue []byte = createAttributeFromTestFile(t, "testattributemultivalue.txt").Value
+func TestInvalidMultiValueNotReturned(t *testing.T) {
+	var nilMultiValueItem = &yotiprotoattr.MultiValue_Value{
```
bucky-boy: Why is this _nilMultiValueItem_?

It looks like an invalid (incomplete) date.  And if the intention is to provide a date that cannot be parsed, you might as well just say 
```
Data:        []byte("invalid")
```

yoti_test.go
```diff
@@ -1212,6 +1234,44 @@ func TestNewImageSlice(t *testing.T) {
 	assertIsExpectedDocumentImagesAttribute(t, documentImagesAttribute.Value(), documentImagesAttribute.Anchors()[0])
 }
 
+func TestImageSliceNotCreatedWithNonMultiValueType(t *testing.T) {
+	attributeName := "test_attribute_name"
```
bucky-boy: If these locals are only used once, maybe don't bother declaring them

yoti_test.go
```diff
@@ -628,6 +628,28 @@ func TestProfile_GetAttribute_String(t *testing.T) {
 	}
 }
 
+func TestEmptyStringIsAllowed(t *testing.T) {
+	attributeValueString := ""
```
bucky-boy: It might read better if its just
```
emptyString := ""
```

attribute/multivalueattribute.go
bucky-boy: Sure.  But why would SDK callers use this method?  Don't we parse everything up front, then just give them the result?

echarrod: Yes, this was a remnant of the the first PR on MultiValue when we were giving clients the option to parse multi values. Now we should handle everything, so that won't be needed, so is not exported.

yoti_test.go
echarrod: Yep good spot, will remove

yoti_test.go
echarrod: Will change to be called invalidMultiValueItem, with the suggested value

PR #38, opened 2019-03-25
SDK-880: Add .travis.yml

echarrod:
- A lot of people seem to be ignoring errors from master, we'll include it for now and see how it goes...
- We have to `go get` the dependencies whilst we are supporting < 1.11.x
- Can add other checks `gometalinter`, `golangci-lint` at a later date, they're not supported out of the box, so we'd have to install them
- go branch: `master` seems to be unreliable, based on the number of other people using it but allowing for failures in it, so I've followed the same approach.
- `windows` is also in beta, and seems to be unreliable, so excluding it for now

[Build Status](https://travis-ci.com/getyoti/yoti-go-sdk)


.travis.yml
```diff
@@ -0,0 +1,45 @@
...
+  # needed until we stop supporting 1.10.x
+  GO111MODULE=off;
+  go get -t -d ./...;   
+
+script:
+  - ./.travis.gofmt.sh
+  - go generate -x ./... && git diff --exit-code; code=$?; git checkout -- .; (exit $code) # Check that go generate ./... produces a zero diff; clean up any changes afterwards.
+  - go vet ./...
+  - go test -v -race ./...
+  - go test -v ./...
```
davidgrayston: Do we need to run tests both with and without `-race`? Can we run only `go test -v -race ./...`?

.travis.yml
```diff
@@ -0,0 +1,45 @@
...
+git:
+  depth: 1
+
+before_script:
+  # needed until we stop supporting 1.10.x
+  GO111MODULE=off;
+  go get -t -d ./...;   
+
+script:
+  - ./.travis.gofmt.sh
```
davidgrayston: Wondering if we need this to be a separate script? Looks like it's common practice to have it separate like this so probably fine.

.travis.yml
echarrod: Looks like the second call is redundant, will remove

.travis.yml
echarrod: I agree it would be nice to have one less file, but seems to be common practice in the google repositories

PR #41, opened 2019-04-23
SDK-895: Undefined content types

echarrod:
After refactoring I got this error on the Travis builds:
```bash
./yoticlient.go:92:32: too many arguments in call to getProfileEndpoint
	have (string, string, string, string)
	want (string, string, string)
```
It looked like there was a clash between the previous version of the `getProfileEndpoint` method, and the current version. 

I ran the Travis docker image locally and looked at this build - 
I could see the updated version of the sdk in `~/getyoti/yoti-go-sdk`, but when I was in `~/gopath/src/github.com/getyoti/yoti-go-sdk`, that's where I could see the old one.

It looks like the old version is being cached in here, so I had to remove the caching. I wouldn't have thought that Go would prioritise looking in the gopath first, when GO111MODULE=on, for versions > 1.10.x, but it looks like it was! 

We can revisit this later, and definitely remove it once 1.10.x has stopped being supported, but it only seems to add a few seconds onto 4 of the builds in the job, so will leave like this for now.

### DIFF
--- a/yoticlient.go
+++ b/yoticlient.go
@@ -86,10 +86,8 @@ func getActivityDetails(requester httpRequester, encryptedToken, sdkID string, k
 		return
 	}
 
-	timestamp := getTimestamp()
-
 	// create http endpoint
-	endpoint := getProfileEndpoint(token, nonce, timestamp, sdkID)
+	endpoint := getProfileEndpoint(token, nonce, sdkID)
 
 	var headers map[string]string
 	if headers, err = createHeaders(key, httpMethod, endpoint, nil); err != nil {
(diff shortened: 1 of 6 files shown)


yoticlient.go
```diff
@@ -104,65 +102,14 @@ func getActivityDetails(requester httpRequester, encryptedToken, sdkID string, k
...
-				addressAttribute := &yotiprotoattr.Attribute{
-					Name:        attrConstAddress,
-					Value:       []byte(formattedAddress),
-					ContentType: yotiprotoattr.ContentType_STRING,
-					Anchors:     protoStructuredPostalAddress.Anchors,
-				}
-
-				profile.attributeSlice = append(profile.attributeSlice, addressAttribute)
-			}
+		return handleSuccessfulResponse(response.Content, key)
```
davidgrayston: Might it be clearer if you assign `userProfile` and `activityDetails` here, as this method always returns `userProfile, activityDetails, errStrings`?

echarrod: As discussed, would be clearer to assign ActivityDetails here, instead of within `handleSuccessfulResponse`, but not worth doing whilst we have the old profile to deal with within that method. Will look at doing that once we do the major release.

PR #44, opened 2019-07-18
SDK-502: New Demo Design

emmas-yoti:


yotiprofile.go
```diff
@@ -116,7 +116,10 @@ func (p Profile) Address() *attribute.StringAttribute {
 	return nil
 }
 
-// StructuredPostalAddress represents the user's address in a JSON format. Will be nil if not provided by Yoti.
+// StructuredPostalAddress represents the user's address in a JSON format.
+// Will be nil if not provided by Yoti. This can be accessed as a
+//	map[string]string{} using a type assertion, e.g.:
```
echarrod: I believe the convention  is to use a space rather than a tab 👍 - https://blog.golang.org/godoc-documenting-go-code
```suggestion
// map[string]string{} using a type assertion, e.g.:
```

yotiprofile.go
```diff
@@ -116,7 +116,10 @@ func (p Profile) Address() *attribute.StringAttribute {
...
+//	structuredPostalAddress := structuredPostalAddressAttribute.Value().(map[string]string{})
```
echarrod:
```suggestion
// structuredPostalAddress := structuredPostalAddressAttribute.Value().(map[string]string{})
```

PR #45, opened 2019-07-19
 SDK-1067: Add simple Error page 

emmas-yoti:
Contains changes from SDK-502 NewDemoDesign, merge that first.

- [x] SDK-502 merged

### DIFF
--- a/examples/profile/main.go
+++ b/examples/profile/main.go
@@ -13,6 +13,7 @@ import (
 	"net/http"
 	"os"
 	"path"
+	"strings"
 
 	yoti "github.com/getyoti/yoti-go-sdk/v2"
 	_ "github.com/joho/godotenv/autoload"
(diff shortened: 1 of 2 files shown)


examples/profile/main.go
```diff
@@ -61,7 +85,13 @@ func profile(w http.ResponseWriter, r *http.Request) {
 
 	activityDetails, errStrings := client.GetActivityDetails(yotiOneTimeUseToken)
 	if len(errStrings) != 0 {
-		log.Fatalf("Errors: %v", errStrings)
+		errorPage(w, r.WithContext(context.WithValue(
+			r.Context(),
+			contextKey("yotiError"),
+			fmt.Sprintf("Errors: %v", errStrings),
```
echarrod: I know it's just using what was previously there, but I don't think we need the `Errors: ` here, since when the error is displayed, we already have `Error: `, so it's `Error: Errors: [ProfileNotFound]`. 

So if we remove "Errors" here, we'll just have `Error: [ProfileNotFound]`

PR #50, opened 2019-07-31
SDK-553: Activity Details Receipt

emmas-yoti:
Dependencies
 - [x] Finish Concourse work (SDK-820 Dev Complete)

### Adding
ApplicationProfile to wrap application attributes
Added baseProfile base class

### Changed
Profile now inherits from baseProfile
Refactored many Profile attributes to use baseProfile.GetStringAttribute to improve DRY
Did the same with GetImageAttribute and GetJSONAttribute.

### Other
I am unsure whether ApplicationProfile needs test coverage as part of this PR. The tests for Profile in yotitest only seem to be concerned with the serialisation/deserialisation of the protoattribute data, none of them seem to be constructing entire Profiles. If there's a counter-example please point it out to me.

This PR going into Development is a Blocker for SDK-822 Create a Dockerised Wrapper Application.


yotiprofile.go
```diff
@@ -164,12 +124,65 @@ func (p Profile) DocumentImages() (*attribute.ImageSliceAttribute, error) {
...
+func (p baseProfile) GetAttribute(attributeName string) *attribute.GenericAttribute {
 	for _, a := range p.attributeSlice {
 		if a.Name == attributeName {
 			return attribute.NewGeneric(a)
 		}
 	}
 	return nil
 }
+
+func (p baseProfile) GetStringAttribute(attributeName string) *attribute.StringAttribute {
```
davidgrayston: Should these methods be exported, or can they be private?

davidgrayston: I missed that this is to allow getting various attribute types by name that don't have a getter - will resolve

yotiprofile.go
```diff
@@ -164,12 +124,65 @@ func (p Profile) DocumentImages() (*attribute.ImageSliceAttribute, error) {
...
+
+// ApplicationLogo is the logo of the application that will be displayed to
+// those users that perform a sharing with it.
+func (p ApplicationProfile) ApplicationLogo() *attribute.ImageAttribute {
+	return p.GetImageAttribute(attrConstApplicationLogo)
+}
+
 // GetAttribute retrieve an attribute by name on the Yoti profile. Will return nil if attribute is not present.
-func (p Profile) GetAttribute(attributeName string) *attribute.GenericAttribute {
+func (p baseProfile) GetAttribute(attributeName string) *attribute.GenericAttribute {
```
echarrod: I _think_ we'll have to think of exposing this method another way - since the previous version was exposed on the profile object, and we're now extending an unexposed struct, it would be technically removing this method from a client's perspective, and so would count as a breaking change.

echarrod:
Oh also, do you think it would be better to split out the BaseProfile and ApplicationProfile functions into separate classes? Just so that class doesn't get too long

yotiprofile.go
emmas-yoti: GetAttribute is still publically exposed on Profile I think? The tests in yoti_test.go still use it.

yotiprofile.go
```diff
@@ -164,12 +124,65 @@ func (p Profile) DocumentImages() (*attribute.ImageSliceAttribute, error) {
 	return nil, nil
 }
 
+// ApplicationName is the name of the application
+func (p ApplicationProfile) ApplicationName() *attribute.StringAttribute {
```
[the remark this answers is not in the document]
emmas-yoti: Tests have been added

yotiprofile.go
echarrod: Yes sorry, you're right, it's still exposed - I just checked by using it from the profile example (I think the tests are in the same package, so it wouldn't need to be exposed in order to access them from there)

emmas-yoti:
Note to self: This branch has gotten tangled with other code changes necessary for Concourse, so I will re-request Review at the end of that block of work to make sure I don't accidentally sneak anything in.

PR #51, opened 2019-08-08
SDK-723: Dynamic Scenarios

emmas-yoti:
### Added
 * DynamicScenario, DynamicScenarioBuilder
 * DynamicPolicy, DynamicPolicyBuilder
 * WantedAttribute, WantedAttributeBuilder
 * Extension, ExtensionBuilder, TransactionalFlow..., LocationConstraint...
 * ShareURL and CreateShareURL

### Changed
Added ClientInterface interface to permit mocking of Yoti Client to test code that depends on it without needing to send real requests.


emmas-yoti:
Just realised I haven't implemented shareURL yet. Oops

emmas-yoti:
ShareURL implemented now.

yoticlient.go
```diff
@@ -63,6 +87,63 @@ func (client *Client) GetActivityDetails(token string) (ActivityDetails, []strin
...
+	return fmt.Errorf(
+		defaultUnknownErrorMessageConst,
+		response.StatusCode,
+		response.Content,
+	)
+}
+
+// MakeRequest is used by other yoti Packages to make requests using a single
+// common client object. Users should not use this function directly
+func (client *Client) MakeRequest(httpMethod, endpoint string, payload []byte, httpErrorMessages ...map[int]string) (responseData string, err error) {
```
davidgrayston: Is it possible to have `GetActivityDetails` and `performAmlCheck` make use of `MakeRequest`?

dynamicsharingservice/policy/dynamic_policy_builder.go
```diff
@@ -0,0 +1,211 @@
+package policy
+
+import (
+	"encoding/json"
+	"strconv"
+)
+
+const (
+	attributeFamilyNameConst              = "family_name"
```
davidgrayston: I can see in other parts of the code we aren't using constants for attributes (like `"family_name"`) - should there be constants defined as per other SDKs to be a bit more DRY? (or is this a limitation in Go?)

emmas-yoti: Should their scope be public or private?

echarrod: Public, see [.NET](https://github.com/getyoti/yoti-dotnet-sdk/blob/development/src/Yoti.Auth/Constants/UserProfile.cs) and [Java](https://github.com/getyoti/yoti-java-sdk/blob/cbe67f434a1326e7b65a54a49474c584a50d628c/yoti-sdk-api/src/main/java/com/yoti/api/attributes/AttributeConstants.java).

it's better for us to share them, rather than having clients having to type them in if they want to request them through Dynamic Scenario or similar 👍

emmas-yoti: I've unified the two attribute name constant lists into one place - it was keeping them private that prevented me doing so earlier.

I can't grep any raw occurences of family_name that aren't going through the constants?

echarrod: Ok, that's fine then! 👍

examples/profile/dynamic-share.html
```diff
@@ -0,0 +1,84 @@
...
+    <section class="yoti-top-section">
+      <div class="yoti-logo-section">
+        <img 
+          class="yoti-logo-image"
+          src="/static/assets/logo.png" 
+          srcset="/static/assets/logo@2x.png 2x"
+          alt="Yoti"/>
+      </div>
+
+      <h1 class="yoti-top-header">We now accept Yoti</h1>
```
echarrod: Could we update this to be "Dynamic Share Example" too? (I know it's the page title. Just would be useful to have more of a distinction between the two pages 👍 
```suggestion
      <h1 class="yoti-top-header">Dynamic Share Example</h1>
```

share_url.go
```diff
@@ -0,0 +1,42 @@
...
+var (
+	// ShareURLHTTPErrorMessages specifies the HTTP error status codes used
+	// by the Share URL API
+	ShareURLHTTPErrorMessages = map[int]string{
+		400: "JSON is incorrect, contains invalid data: %[2]s",
+		404: "Application was not found: %[2]s",
+	}
+)
+
+// ShareURL contains a dynamic share qr code
```
echarrod: Same here with consistent capitalisation. My preference would be `QR code`
```suggestion
// ShareURL contains a dynamic share QR code
```

yoticlient.go
```diff
@@ -70,73 +95,114 @@ func (client *Client) getAPIURL() string {
...
+	}
+	return fmt.Errorf(
+		defaultUnknownErrorMessageConst,
+		response.StatusCode,
+		response.Content,
+	)
+}
+
+// MakeRequest is used by other yoti Packages to make requests using a single
+// common client object. Users should not use this function directly
```
echarrod: If this isn't exported, do we need to have this comment?

dynamic_scenario_builder.go
```diff
@@ -0,0 +1,61 @@
...
+	builder.scenario.callbackEndpoint = endpoint
+	return builder
+}
+
+// Build constructs the DynamicScenario
+func (builder *DynamicScenarioBuilder) Build() DynamicScenario {
+	return builder.scenario
+}
+
+// MarshalJSON ...
```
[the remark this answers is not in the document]
emmas-yoti: Some linters (specifically the one build into vim-go) complains if you have an exported function without a doc comment, but the MarshalJSON function is built into how encoding/json json.Marshal works. I couldn't think of a good boilerplate comment to put here.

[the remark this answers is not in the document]
echarrod: Yes we should always have comments on exported functions. 

I think we should have always have some text, even if it can seem redundant. We could use something like 
> // MarshalJSON returns m as the JSON encoding of m.

(From https://golang.google.cn/src/encoding/json/stream.go - L256)

extension/location_constraint_extension.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+// WithLatitude sets the latitude of the location constraint
+func (builder *LocationConstraintExtensionBuilder) WithLatitude(latitude float64) *LocationConstraintExtensionBuilder {
+	builder.extension.latitude = latitude
+	return builder
+}
+
+// WithLongitude sets the longitude of the location constraint
+func (builder *LocationConstraintExtensionBuilder) WithLongtitude(longitude float64) *LocationConstraintExtensionBuilder {
```
davidgrayston: The `WithLongtitude` function has the extra "t"

PR #57, opened 2019-09-02
Sdk 820 anchor extension scanning

emmas-yoti:
### Dependency
 * [ ] SDK-723 Dynamic Scenarios

### Changed
Anchor extension scanning now stashes the last extension match and uses that to set the anchor type instead of using the last extension seen.


anchor/anchorparser.go
```diff
@@ -37,13 +37,14 @@ func ParseAnchors(protoAnchors []*yotiprotoattr.Anchor) []*Anchor {
 					value string
 					err   error
 				)
-				anchorType, value, err = parseExtension(ext)
+				parsedAnchorType, value, err := parseExtension(ext)
```
MrBurtyyy: I can't remember 100%, but I think we need to consider the case of parsing actual Unknown anchors.  I think this currently would just abandon an unknown anchor (instead of just the critical extension).  Do we have any tests for unknown anchors in Go?

emmas-yoti: TestAnchorParser_UnknownAnchor (yoti_test.go:973)

PR #61, opened 2019-09-16
Fix order of imports issue causing code analysis to fail on development

emmas-yoti:


emmas-yoti:
> Doesn't one of our precommit hooks reorder the imports? It won't revert this back to how it was before will it?

It's the same algorithm/util generating the error (goimports). I think there might be an identical patch in the precommit branch though

PR #62, opened 2019-09-17
Release 2.6.0

emmas-yoti:
### Added
 * ActivityDetails:
   * `ApplicationProfile`
   * `Timestamp()`
   * `ReceiptID()`
 * Profile:
   * `DocumentDetails()`
 * Requesting dynamic shares: `yoti.CreateShareURL(&client, &scenario)`
   * Including Source-based constraints
 * Signed request builder: `SignedRequest`
### Changed
 * Anchor timestamp now include Microsecond precision
* Improve Profile example project


yotiprofile.go
```diff
@@ -2,104 +2,77 @@ package yoti
...
+	AttrConstMobileNumber              = "phone_number"
+	AttrConstEmailAddress              = "email_address"
+	AttrConstDateOfBirth               = "date_of_birth"
+	AttrConstAddress                   = "postal_address"
+	AttrConstStructuredPostalAddress   = "structured_postal_address"
+	AttrConstGender                    = "gender"
+	AttrConstNationality               = "nationality"
+	AttrConstDocumentImages            = "document_images"
+	AttrConstDocumentDetails           = "document_details"
+	AttrConstApplicationName           = "application_name"
```
echarrod: Would it make more sense for the ApplicationProfile constants to live in `application_profile.go` rather than here?

emmas-yoti:
> Also do you think it would be wise to put the dynamic scenarios files in their own subpackage? It seems like there's at least 3 files and 3 accompanying test files that I think we can do that for.

Unfortunately this would introduce a circular dependency between dynamic_policy and the AttrConst* constants in yotiprofile/applicationprofile.

I can't think of a solution to this that wouldn't either be a breaking change or break DRY. In version 3 we could add a 'constants' subpackage and move AttrConst* to there.

examples/profile/main.go
```diff
@@ -48,6 +48,104 @@ func home(w http.ResponseWriter, req *http.Request) {
...
+		)))
+		return
+	}
+
+	templateVars := map[string]interface{}{
+		"yotiClientSdkID": sdkID,
+		"yotiShareURL":    share.ShareURL,
+	}
+
+	t, err := template.ParseFiles("dynamic-share.html")
```
echarrod: Do you think we should add a source-constrains.html file too? Or maybe it would be better to change the title and header to indicate that it's different to the standard dynamic share page?

examples/profile/main.go
```diff
@@ -48,6 +48,104 @@ func home(w http.ResponseWriter, req *http.Request) {
...
+	if err != nil {
+		panic("Error parsing template: " + err.Error())
+	}
+
+	err = t.Execute(w, templateVars)
+	if err != nil {
+		panic("Error applying the parsed template: " + err.Error())
+	}
+}
+func dynamicShare(w http.ResponseWriter, req *http.Request) {
```
echarrod: Looks like this is almost identical to the sourceConstraints method above, can we extract a lot of the functionality into a common method?

yotiprotoattr/Attribute.pb.go
```diff
@@ -5,8 +5,9 @@ package yotiprotoattr
 
 import (
 	fmt "fmt"
-	proto "github.com/golang/protobuf/proto"
 	math "math"
+
```
echarrod: Sorry, I should have noticed this before - since the protobuf files are all auto-generated, we have generally taken a decision to ignore them from all linting, since we'd have to update them again every time we regenerate, and it would add noise to the changeset every time.

emmas-yoti: Oops. Not sure when this snuck in. I don't think there's anything I can do about this now.

PR #71, opened 2019-11-06
SDK-1147: Add support for multiple attributes with the same name

MrBurtyyy:
## Added

* `BaseProfile.GetAttributes()`


baseprofile.go
```diff
@@ -19,6 +19,17 @@ func (p baseProfile) GetAttribute(attributeName string) *attribute.GenericAttrib
...
+func (p baseProfile) GetAttributes(attributeName string) []*attribute.GenericAttribute {
+	var attributes []*attribute.GenericAttribute
+	for _, a := range p.attributeSlice {
+		if a.Name == attributeName {
+			attributes = append(attributes, attribute.NewGeneric(a))
+		}
+	}
+	return attributes
+}
+
```
davidgrayston: Could we have a method for each type, e.g. `GetStringAttributes()`?

echarrod: I think since most clients won't be using this, I'm not a huge fan of adding a method for each of the 7 Content Types we support. We can add this for now, and then if things change (or Go gets generics(!)), then we can look at changing it 👍

echarrod: From the tests, it doesn't seem too strenuous to cast it, so I think we're fine:
```go
assert.Assert(t, is.Equal(fullNames[0].Value().(string), "some_value"))
```

jveling-yoti (approved):

PR #74, opened 2019-11-19
SDK-1275: Change ActivityDetails errstrings to provide something more meaningful

### LINKED ISSUE #73: Are errStrings defined?
> ```go
> activityDetails, errStrings := client.GetActivityDetails(yotiOneTimeUseToken)
> if len(errStrings) != 0 {
>   // handle unhappy path
> }
> ```

The example above, given in the readme implies that errStrings can be handled, but there is no documentation listing what error strings can be returned.

Without knowing what errStrings can be returned callers of `GetActivityDetails` will be unable to write any code to handle the unhappy path aside from just failing. This prevents them from giving any feedback to callers that they in turn could react to. Notably they will be unable to determine if the error returned was as a consequence of bad input or if it was caused by a temporary server failure and are able to retry.

emmas-yoti:
### Additions
 * Add error types in `errors.go` and `requests/errors.go` which can be used for flow control by client code.

### Changes
 * Change `errStrings []string` in `GetActivityDetails` to `err error`
 * Use `HTTPError` to describe HTTP-Request based error messages rather than using strings

resolves #73


errors.go
```diff
@@ -0,0 +1,69 @@
...
+	return "Invalid Private Key"
+}
+
+// Unwrap obtains the wrapped error
+func (err InvalidKeyError) Unwrap() error {
+	return err.nestedError
+}
+
+// HTTPError describes an error returned from the remote HTTP/REST API
+type HTTPError struct {
```
piersy: I would advise against having an `HTTPError` unless it is for internal use (in which case it should be private). Because it will require users of the SDK to understand what http error codes can be returned from services that the sdk interacts with, which puts more burden on them.

Also isn't it intended that the SDK abstracts the internals of the Yoti services away from it's users?

yoticlient.go
```diff
@@ -100,22 +99,22 @@ func (client *Client) GetSdkID() string {
...
 }
 
-func (client *Client) getActivityDetails(token string) (userProfile UserProfile, activity ActivityDetails, errStrings []string) {
+func (client *Client) getActivityDetails(token string) (userProfile UserProfile, activity ActivityDetails, err error) {
 
 	httpMethod := http.MethodGet
 	key, err := loadRsaKey(client.Key)
 	if err != nil {
-		errStrings = append(errStrings, fmt.Sprintf("Invalid Key: %s", err.Error()))
+		err = InvalidKeyError{nestedError: err}
```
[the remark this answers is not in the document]
emmas-yoti: I agree that's better but:
1) We've previously expressed a desire to keep all the APIs as similar as possible to make maintenance as easy as possible - so if we decide to change it here, we need to have a plan to change the behavior in every other SDK to match.

2) It's beyond the scope of work for this ticket.

So I think we should make a separate Story for it with sub-tasks for each SDK.

errors.go
emmas-yoti: So we discussed this offline:

I was given the impression that part of this work was that we had a customer who wanted a list of error values for them to handle (how I wasn't clear) in their business logic. We already provide a success/fail value where err != nil - with a string that can be used by support for debugging/logging but one that can't be programmatically compared for flow control (go compares errors by reference, not value).

Ed suggested having a meeting about what we want the requirements to be when everyone's in the office (Tuesday maybe?)

piersy: Sounds good :+1:

yoticlient.go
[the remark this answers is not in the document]
piersy: Fair enough :+1:

piersy:
Hi @emmas-yoti just adding the definition of `temp.Error` that we came up with the other day.

```go
package temp

// Error is a temporary error. Trying the same operation again with the same
// parameters might change the result. This error is useful when the result of
// an operation depends on external factors that might change over time.
// E.G Trying to access an external API through a network connection could
// result in a temporary error if the service is down. Retrying again with the
// same parameters could work if the external service gets fixed.
type Error struct {
	Err error
}

func (e *Error) Error() string {
	return e.Err.Error()
}

func (e *Error) Unwrap() error {
	return e.Err
}
``` 

To mark an error as temp you simply do
```go
err = &temp.Error{err}
```

yoticlient.go
```diff
@@ -222,26 +220,21 @@ func getProtobufAttribute(profile Profile, key string) *yotiprotoattr.Attribute
...
 	var parsedResponse = profileDO{}
-	var err error
 
 	if err = json.Unmarshal([]byte(responseContent), &parsedResponse); err != nil {
-		errStrings = append(errStrings, err.Error())
 		return
 	}
 
 	if parsedResponse.Receipt.SharingOutcome != "SUCCESS" {
 		err = ErrSharingFailure
```
emmas-yoti: I couldn't actually find a code path where ErrStrings.length >1. Reasoning will be presented in comments...

ErrStrings was null at this point, so we don't need a list here.

yoticlient.go
```diff
@@ -222,26 +220,21 @@ func getProtobufAttribute(profile Profile, key string) *yotiprotoattr.Attribute
...
-		errStrings = append(errStrings, err.Error())
 	} else {
 		var attributeList, appAttributeList *yotiprotoattr.AttributeList
 		if attributeList, err = decryptCurrentUserReceipt(&parsedResponse.Receipt, key); err != nil {
-			errStrings = append(errStrings, err.Error())
 			return
```
emmas-yoti: The previous assignment happened on the other branch of an if statement, so errStrings was null here... then we immediately return so nothing else can be appended. So errStrings can only be length 1

yoticlient.go
```diff
@@ -222,26 +220,21 @@ func getProtobufAttribute(profile Profile, key string) *yotiprotoattr.Attribute
...
 		}
 		if appAttributeList, err = decryptCurrentApplicationProfile(&parsedResponse.Receipt, key); err != nil {
-			errStrings = append(errStrings, err.Error())
```
emmas-yoti: errStrings was null before this. And then we immediately return, so length is again 1

yoticlient.go
```diff
@@ -265,7 +258,6 @@ func handleSuccessfulResponse(responseContent string, key *rsa.PrivateKey) (user
 			log.Printf("Unable to get 'Formatted Address' from 'Structured Postal Address'. Error: %q", err)
 		} else if formattedAddress != "" {
 			if _, err = profile.StructuredPostalAddress(); err != nil {
-				errStrings = append(errStrings, err.Error())
```
emmas-yoti: errStrings was null before this. And then we immediately return, so length is again 1

yoticlient.go
```diff
@@ -291,7 +283,7 @@ func handleSuccessfulResponse(responseContent string, key *rsa.PrivateKey) (user
 		}
 	}
 
-	return userProfile, activityDetails, errStrings
```
emmas-yoti: Length is 1 or 0 here.

yoticlient.go
```diff
@@ -100,22 +95,22 @@ func (client *Client) GetSdkID() string {
...
+	return activity, err
 }
 
-func (client *Client) getActivityDetails(token string) (userProfile UserProfile, activity ActivityDetails, errStrings []string) {
+func (client *Client) getActivityDetails(token string) (userProfile UserProfile, activity ActivityDetails, err error) {
 
 	httpMethod := http.MethodGet
 	key, err := loadRsaKey(client.Key)
 	if err != nil {
-		errStrings = append(errStrings, fmt.Sprintf("Invalid Key: %s", err.Error()))
```
emmas-yoti: errStrings was nil before this point, and we immediately return - so it can never be more than length 1

yoticlient.go
```diff
@@ -100,22 +95,22 @@ func (client *Client) GetSdkID() string {
...
+		err = fmt.Errorf("Invalid Key: %s", err.Error())
 		return
 	}
 	token, err = decryptToken(token, key)
 	if err != nil {
-		errStrings = append(errStrings, fmt.Sprintf("Invalid Key: %s", err.Error()))
```
emmas-yoti: Ditto here.

yoticlient.go
```diff
@@ -129,7 +124,6 @@ func (client *Client) getActivityDetails(token string) (userProfile UserProfile,
 		DefaultHTTPErrorMessages,
 	)
 	if err != nil {
-		errStrings = append(errStrings, err.Error())
```
emmas-yoti: Ditto here.

README.md
```diff
@@ -110,12 +110,28 @@ Keeping your settings and access keys outside your repository is highly recommen
...
-activityDetails, errStrings := client.GetActivityDetails(yotiOneTimeUseToken)
-if len(errStrings) != 0 {
+activityDetails, err := client.GetActivityDetails(yotiOneTimeUseToken)
+if err != nil {
   // handle unhappy path
 }
 ```
 
+## Handling Errors
+If a network error occurs that can be handled by resending the request occurs,
```
echarrod:
```suggestion
If a network error occurs that can be handled by resending the request,
```

README.md
```diff
@@ -110,12 +110,28 @@ Keeping your settings and access keys outside your repository is highly recommen
...
+the error returned by the SDK will have the TemporaryError type. This can be tested
+for using either errors.Is or a type assertion and resent.
```
echarrod: Just would be good to wrap this in quotes so it's clear it's not a full stop:
```suggestion
for using either `errors.Is`, or a type assertion, and resent.
```

yoticlient.go
```diff
@@ -100,22 +95,22 @@ func (client *Client) GetSdkID() string {
 // GetActivityDetails requests information about a Yoti user using the one time use token generated by the Yoti login process.
 // It returns the outcome of the request. If the request was successful it will include the users details, otherwise
 // it will specify a reason the request failed.
```
piersy:
```suggestion
// an error will be returned which will specify the reason the request failed.
```

yoticlient.go
```diff
@@ -100,22 +95,22 @@ func (client *Client) GetSdkID() string {
 // GetActivityDetails requests information about a Yoti user using the one time use token generated by the Yoti login process.
```
piersy: It would be nice to mention here that this function may return `yoti.TemporaryErrors`

errors.go
```diff
@@ -0,0 +1,16 @@
+package yoti
+
+// TemporaryError indicates that a temporary outage has occured and the
```
piersy: I have just realised that this is not a good approach to marking temporary errors.

This is because I hadn't realised that you can use As to check for interfaces as well as concrete error structs. The std go packages already use an interface with method `Temporary() bool` that they use to indicate temporary errors. We should follow their approach and add a `Temporary` method to this error that always returns true.

This is an example of using `As` to check for an interface - https://play.golang.org/p/IVjEsfflbNJ

This means we can stop wrapping errors from `doRequest` since go will have already marked them as `Temporary` but we can still use this struct to wrap http errors for 500 status codes.

The documentation will need to be updated to show checking for `Temporary` errors and we should remove reference to the yoti.TemoraryError.

yoticlient.go
```diff
@@ -99,23 +94,25 @@ func (client *Client) GetSdkID() string {
 
 // GetActivityDetails requests information about a Yoti user using the one time use token generated by the Yoti login process.
 // It returns the outcome of the request. If the request was successful it will include the users details, otherwise
-// it will specify a reason the request failed.
-func (client *Client) GetActivityDetails(token string) (ActivityDetails, []string) {
-	_, activity, errStrings := client.getActivityDetails(token)
-	return activity, errStrings
+// an error it will be returned which will specify the reason the request failed,
```
echarrod:
```suggestion
// an error will be returned, which will specify the reason the request failed.
```
On top of this, it looks like the two lines above are longer, and this one and the one below are shorter. 
Don't mind which one we go for (slight preference on shorter lines), but can we make them all the same 👍

yoticlient.go
```diff
@@ -99,23 +94,25 @@ func (client *Client) GetSdkID() string {
 
 // GetActivityDetails requests information about a Yoti user using the one time use token generated by the Yoti login process.
 // It returns the outcome of the request. If the request was successful it will include the users details, otherwise
```
echarrod: I know you didn't change this, but I just noticed it
```suggestion
 // It returns the outcome of the request. If the request was successful it will include the user's details, otherwise
```

yoticlient.go
```diff
@@ -99,23 +94,25 @@ func (client *Client) GetSdkID() string {
...
-// it will specify a reason the request failed.
-func (client *Client) GetActivityDetails(token string) (ActivityDetails, []string) {
-	_, activity, errStrings := client.getActivityDetails(token)
-	return activity, errStrings
+// an error it will be returned which will specify the reason the request failed,
+// if the function call can be reattempted with the same token the error will
```
echarrod: Just need to do this change to if you decide to accept my last suggestion of adding a full stop.
```suggestion
// If the function call can be reattempted with the same token the error will
```

yoti_test.go
```diff
@@ -72,10 +72,12 @@ func TestYotiClient_HttpFailure_ReturnsFailure(t *testing.T) {
...
 
-	_, _, errorStrings := client.getActivityDetails(encryptedToken)
+	_, _, err := client.getActivityDetails(encryptedToken)
 
-	assert.Assert(t, len(errorStrings) > 0)
-	assert.Check(t, strings.HasPrefix(errorStrings[0], "Unknown HTTP Error"))
+	assert.Check(t, err != nil)
+	assert.Check(t, strings.HasPrefix(err.Error(), "Unknown HTTP Error"))
+	_, temporary := err.(TemporaryError)
+	assert.Check(t, temporary)
```
piersy: The tests would be more flexible here if you check for `interface{ Temporary() bool}` you would also wan't to check the return of Temporary.

yoti_test.go
```diff
@@ -52,10 +52,10 @@ func TestYotiClient_KeyLoad_Failure(t *testing.T) {
 		},
 	}
 
-	_, _, errorStrings := client.getActivityDetails(encryptedToken)
+	_, _, err := client.getActivityDetails(encryptedToken)
 
-	assert.Assert(t, len(errorStrings) > 0)
-	assert.Check(t, strings.HasPrefix(errorStrings[0], "Invalid Key"))
+	assert.Check(t, err != nil)
+	assert.Check(t, strings.HasPrefix(err.Error(), "Invalid Key"))
```
piersy: Would be good to check that this error is not marked as temporary.

yoti_test.go
```diff
@@ -92,10 +94,10 @@ func TestYotiClient_HttpFailure_ReturnsProfileNotFound(t *testing.T) {
 		},
 	}
 
-	_, _, errorStrings := client.getActivityDetails(encryptedToken)
+	_, _, err := client.getActivityDetails(encryptedToken)
 
-	assert.Assert(t, len(errorStrings) > 0)
-	assert.Check(t, strings.HasPrefix(errorStrings[0], "Profile Not Found"))
+	assert.Check(t, err != nil)
+	assert.Check(t, strings.HasPrefix(err.Error(), "Profile Not Found"))
```
piersy: In fact in all places where you are checking errors in tests it would be good to assert the temporaryness of them.

yoticlient.go
```diff
@@ -97,25 +92,28 @@ func (client *Client) GetSdkID() string {
...
-// it will specify a reason the request failed.
-func (client *Client) GetActivityDetails(token string) (ActivityDetails, []string) {
-	_, activity, errStrings := client.getActivityDetails(token)
-	return activity, errStrings
+// GetActivityDetails requests information about a Yoti user using the one time
+// use token generated by the Yoti login process. It returns the outcome of the
+// request. If the request was successful it will include the user's details,
+// otherwise an error will be returned, which will specify the reason the
+// request failed. If the function call can be reattempted with the same token
+// the error will be a temporary error.
```
piersy:
```suggestion
// the error will implement interface{ Temporary() bool }.
```

yoticlient.go
```diff
@@ -97,25 +92,28 @@ func (client *Client) GetSdkID() string {
...
 	key, err := loadRsaKey(client.Key)
 	if err != nil {
-		errStrings = append(errStrings, fmt.Sprintf("Invalid Key: %s", err.Error()))
+		err = fmt.Errorf("Invalid Key: %s", err.Error())
 		return
 	}
 	token, err = decryptToken(token, key)
 	if err != nil {
-		errStrings = append(errStrings, fmt.Sprintf("Invalid Key: %s", err.Error()))
+		err = fmt.Errorf("Invalid Key: %s", err.Error())
```
piersy: Invalid Key looks like the wrong error message here.

It would also be good to have a test that checks this error case, and I guess it would assert that the error is not temporary, since this is a failure to decrypt a token which feels pretty permanent.

yoti_test.go
```diff
@@ -52,10 +52,40 @@ func TestYotiClient_KeyLoad_Failure(t *testing.T) {
...
+	_, _, err := client.getActivityDetails(encryptedToken)
 
-	assert.Assert(t, len(errorStrings) > 0)
-	assert.Check(t, strings.HasPrefix(errorStrings[0], "Invalid Key"))
+	assert.Check(t, err != nil)
+	assert.Check(t, strings.HasPrefix(err.Error(), "Invalid Key"))
+	type Interface interface {
+		Temporary() bool
+	}
+	_, temporary := err.(Interface)
```
piersy: You could represent this like so

```go
_, temporary := err.(interface{Temporary() bool})
```

Which just seems a bit neater to me.

Also we should be checking the value returned by Temporary on the temporary error since it is possible for Temporary to return false in the case of Temporary errors returned by the go standard libs.

yoti_test.go
```diff
@@ -52,10 +52,34 @@ func TestYotiClient_KeyLoad_Failure(t *testing.T) {
...
+			},
+		},
+	}
+
+	_, _, err := client.getActivityDetails("")
+
+	assert.Check(t, err != nil)
+	assert.Check(t, strings.HasPrefix(err.Error(), "Invalid Token"))
+	tempError, temporary := err.(interface{ Temporary() bool })
+	assert.Check(t, !temporary || tempError.Temporary())
```
piersy:
```suggestion
	assert.Check(t, !temporary || !tempError.Temporary())
```

yoti_test.go
```diff
@@ -52,10 +52,34 @@ func TestYotiClient_KeyLoad_Failure(t *testing.T) {
...
 
-	_, _, errorStrings := client.getActivityDetails(encryptedToken)
+	_, _, err := client.getActivityDetails(encryptedToken)
 
-	assert.Assert(t, len(errorStrings) > 0)
-	assert.Check(t, strings.HasPrefix(errorStrings[0], "Invalid Key"))
+	assert.Check(t, err != nil)
+	assert.Check(t, strings.HasPrefix(err.Error(), "Invalid Key"))
+	tempError, temporary := err.(interface{ Temporary() bool })
+	assert.Check(t, !temporary || tempError.Temporary())
```
piersy:
```suggestion
	assert.Check(t, !temporary || !tempError.Temporary())
```

yoti_test.go
```diff
@@ -92,10 +119,12 @@ func TestYotiClient_HttpFailure_ReturnsProfileNotFound(t *testing.T) {
...
 
-	_, _, errorStrings := client.getActivityDetails(encryptedToken)
+	_, _, err := client.getActivityDetails(encryptedToken)
 
-	assert.Assert(t, len(errorStrings) > 0)
-	assert.Check(t, strings.HasPrefix(errorStrings[0], "Profile Not Found"))
+	assert.Check(t, err != nil)
+	assert.Check(t, strings.HasPrefix(err.Error(), "Profile Not Found"))
+	tempError, temporary := err.(interface{ Temporary() bool })
+	assert.Check(t, !temporary || tempError.Temporary())
```
piersy:
```suggestion
	assert.Check(t, !temporary || !tempError.Temporary())
```

yoti_test.go
```diff
@@ -113,10 +142,12 @@ func TestYotiClient_SharingFailure_ReturnsFailure(t *testing.T) {
...
 
-	_, _, errorStrings := client.getActivityDetails(encryptedToken)
+	_, _, err := client.getActivityDetails(encryptedToken)
 
-	assert.Assert(t, len(errorStrings) > 0)
-	assert.Check(t, strings.HasPrefix(errorStrings[0], ErrSharingFailure.Error()))
+	assert.Check(t, err != nil)
+	assert.Check(t, strings.HasPrefix(err.Error(), ErrSharingFailure.Error()))
+	tempError, temporary := err.(interface{ Temporary() bool })
+	assert.Check(t, !temporary || tempError.Temporary())
```
piersy:
```suggestion
	assert.Check(t, !temporary || !tempError.Temporary())
```

README.md
```diff
@@ -110,12 +110,28 @@ Keeping your settings and access keys outside your repository is highly recommen
...
+activityDetails, err := client.GetActivityDetails(yotiOneTimeUseToken)
+if err != nil {
   // handle unhappy path
 }
 ```
 
+## Handling Errors
+If a network error occurs that can be handled by resending the request,
+the error returned by the SDK will implement the temporary error interface. 
+This can be tested for using either `errors.Is` or a type assertion, and resent.
```
davidgrayston:
```suggestion
This can be tested for using either `errors.Is` or a type assertion, and resend.
```

PR #76, opened 2019-11-22
SDK-1283: Prevent Empty Wanted Attribute Names

emmas-yoti:
### Additions
 * Add `validate.go` with `notEmpty` to verify a string is not empty
 * `WantedAttribute.MarshalJSON` returns an error message if `attr.name` is empty

### DIFF
--- a/wantedattributebuilder.go
+++ b/wantedattributebuilder.go
@@ -65,6 +65,10 @@ func (builder *WantedAttributeBuilder) Build() WantedAttribute {
 
 // MarshalJSON returns the JSON encoding
 func (attr *WantedAttribute) MarshalJSON() ([]byte, error) {
+	err := notEmpty(attr.name, "Wanted attribute names must not be empty")
+	if err != nil {
+		return nil, err
+	}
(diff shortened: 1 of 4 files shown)


wantedattributebuilder.go
```diff
@@ -65,6 +65,10 @@ func (builder *WantedAttributeBuilder) Build() WantedAttribute {
 
 // MarshalJSON returns the JSON encoding
 func (attr *WantedAttribute) MarshalJSON() ([]byte, error) {
+	err := notEmpty(attr.name, "Wanted attribute names must not be empty")
```
echarrod: We can put this validation in the `.Build()` function can't we, to catch it earlier?

emmas-yoti: The Build function doesn't return an error - so there's no mechanism to communicate the error back to the caller (without a breaking change)

echarrod: Ah. We'll leave it like this for now. We can revisit whether to move the error to `Build()` in the major (I think we probably should)

PR #83, opened 2019-12-09
SDK-1292: Load PEM file as part of client initialisation

emmas-yoti:
Change type of Key in Client. Add LoadPEM method for loading Key from
PEM bytes

### DIFF
--- a/examples/aml/main.go
+++ b/examples/aml/main.go
@@ -26,8 +26,7 @@ func main() {
 		return
 	}
 
-	rsaKey, err := yoti.LoadPEM(key)
-	client := yoti.NewClient(sdkID, rsaKey)
+	client, err := yoti.NewClient(sdkID, key)
 	if err != nil {
 		log.Printf("Problem initialising client: Error: `%s`", err)
 		return
--- a/examples/aml/main.go
+++ b/examples/aml/main.go
@@ -28,7 +28,12 @@ func main() {
 
 	client = &yoti.Client{
 		SdkID: sdkID,
-		Key:   key}
+	}
+	client.Key, err = yoti.LoadPEM(key)
+	if err != nil {
+		log.Printf("Problem initialising client: Error: `%s`", err)
+		return
+	}
 
 	givenNames := "Edward Richard George"
 	familyName := "Heath"
--- a/examples/aml/main.go
+++ b/examples/aml/main.go
@@ -26,10 +26,8 @@ func main() {
 		return
 	}
 
-	client = &yoti.Client{
-		SdkID: sdkID,
-	}
-	client.Key, err = yoti.LoadPEM(key)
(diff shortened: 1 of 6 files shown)


examples/aml/main.go
```diff
@@ -28,7 +28,8 @@ func main() {
 
 	client = &yoti.Client{
 		SdkID: sdkID,
-		Key:   key}
+	}
+	client.LoadPEM(key)
```
davidgrayston: Would it be clearer to provide the key like this?
```go
key, err := yoti.LoadPEM(source)

client = &yoti.Client{
    SdkID: sdkID,
    Key:  key
}
```

echarrod: I think it would be better to do this and be explicit about what the function is doing. Although we might want to call it a new name actually if clients are passing `key` into the struct as before, but the type is different

examples/profile/main.go
```diff
@@ -144,7 +151,15 @@ func profile(w http.ResponseWriter, r *http.Request) {
 
 	client = &yoti.Client{
 		SdkID: sdkID,
-		Key:   key}
+	}
+	client.Key, err = yoti.LoadPEM(key)
```
echarrod: I was thinking we would still have people initialise the client in the same way as before, and we just have the loading of the pem file into a `key` variable above it. 
```go
rsaPrivateKey, keyErr := yoti.LoadPEM(key)
if keyErr != nil {
	errorPage(w, r.WithContext(context.WithValue(
		r.Context(),
		contextKey("yotiError"),
		fmt.Sprintf("%s", keyErr),
	)))
}

client = &yoti.Client{
	SdkID: sdkID,
	Key:   rsaPrivateKey,
}
```
Then it's more clear that it is a required part of initialisation. If we have it like this, it might allow, or just suggest, that clients can initialise the client without a key.

emmas-yoti: We need to return and check for an error value.

yoti_test.go
```diff
@@ -262,10 +260,13 @@ func TestYotiClient_ParseWithoutProfile_Success(t *testing.T) {
 				},
 			},
 		}
+		var err error
+		client.Key, err = loadRsaKey(key)
+		assert.NilError(t, err)
 
-		userProfile, activityDetails, err := client.getActivityDetails(encryptedToken)
+		userProfile, activityDetails, errStrings := client.getActivityDetails(encryptedToken)
```
echarrod: Don't think the `errStrings` changes were meant to be included, right?

examples/profile/main.go
emmas-yoti: Have discussed how we want to handle this offline. Have made those changes.

yoti_test.go
emmas-yoti: I'll have to manually fix the merge conflict one way or the other.

PR #84, opened 2019-12-10
SDK-1252: Make embedded yotiprotoattr attribute a private member

emmas-yoti:
Add Name() and ContentType() methods to expose functionality previously
accessed through the yotiprotoattr.Attribute interface

### DIFF
--- a/yoti_test.go
+++ b/yoti_test.go
@@ -464,7 +464,7 @@ func TestYotiClient_MissingPostalAddress_UsesFormattedAddress(t *testing.T) {
 	structuredPostalAddress, err = profile.StructuredPostalAddress()
 
 	assert.Assert(t, is.Nil(err))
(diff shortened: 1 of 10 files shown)


yoti_test.go
```diff
@@ -464,7 +464,7 @@ func TestYotiClient_MissingPostalAddress_UsesFormattedAddress(t *testing.T) {
 	structuredPostalAddress, err = profile.StructuredPostalAddress()
 
 	assert.Assert(t, is.Nil(err))
-	assert.Equal(t, structuredPostalAddress.ContentType, yotiprotoattr.ContentType_JSON, "Retrieved attribute does not have the correct type")
+	assert.Equal(t, structuredPostalAddress.ContentType(), yotiprotoattr.ContentType_JSON.String(), "Retrieved attribute does not have the correct type")
```
echarrod: So there's no way we can prevent clients from having to add `()` to their code, if we also want to keep these values read-only, right?

emmas-yoti: Yup.

PR #86, opened 2019-12-11
SDK-490: Remove old profile object

emmas-yoti:


yoti_test.go
```diff
@@ -452,67 +433,9 @@ func TestYotiClient_MissingPostalAddress_UsesFormattedAddress(t *testing.T) {
...
-		ID:                      "remember_me_id0123456789",
-		OtherAttributes:         make(map[string]AttributeValue),
-		StructuredPostalAddress: structuredAddress,
-		Address:                 ""}
-
-	address, err := ensureAddressUserProfile(result)
-
-	assert.Assert(t, is.Nil(err), "Failed to add formatted address to address")
-	assert.Equal(t, address, "", "Formatted address missing, but address was still changed")
 }
```
echarrod: Looks like we've removed the only tests which covered lines 247-261 in yoticlient.go (replacing postal address with formatted address, if the former is missing and the latter is present).

Can we add tests for this?

Would be good to get our automated coverage checks added to Go so we can automatically flag things like this too

echarrod: Looks like it was my mistake, those lines weren't covered before. Please would you be able to add a ticket for adding coverage for those lines though (or just adding coverage for that file in general)?

PR #87, opened 2019-12-12
SDK-1322: Add dependabot configuration file

MrBurtyyy:
## Added

* Dependabot configuration file to change default behaviour

### DIFF
--- a/.dependabot/config.yml
+++ b/.dependabot/config.yml
@@ -2,7 +2,7 @@ version: 1
 update_configs:
   - package_manager: "go:modules"
     directory: "/"
-    update_schedule: "live"
+    update_schedule: "daily"
     target_branch: "development"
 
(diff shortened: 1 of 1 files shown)


.dependabot/config.yml
```diff
@@ -0,0 +1,12 @@
+version: 1
+update_configs:
+  - package_manager: "go:modules"
+    directory: "/"
+    update_schedule: "live"
```
echarrod:
```suggestion
    update_schedule: "daily"
```
Unfortunately it looks like the only options available for Go are  daily, weekly, monthly.

When I validate the file with https://dependabot.com/docs/config-file/validator/ I get:
> **Validation errors** The property '#/update_configs/0/update_schedule' value "live" did not match one of the following values: daily, weekly, monthly

PR #88, opened 2019-12-13
SDK-849: Change how generic attribute parsing is handled in Go

emmas-yoti:
### Changed
 * `attribute.parseValue` parsed image types to `*attribute.Image`
 * `ContentType_UNDEFINED` explicitly parses attribute data to a string, without emitting warning logging

### DIFF
--- a/attribute/parser.go
+++ b/attribute/parser.go
@@ -46,7 +46,7 @@ func parseValue(contentType yotiprotoattr.ContentType, byteValue []byte) (interf
 
 	case yotiprotoattr.ContentType_JPEG,
 		yotiprotoattr.ContentType_PNG:
-		return byteValue, nil
+		return ParseImageValue(contentType, byteValue)
 
 	default:
 		log.Printf("Unknown type '%s', attempting to parse it as a String", contentType)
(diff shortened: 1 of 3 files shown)


attribute/parser.go
```diff
@@ -46,7 +46,10 @@ func parseValue(contentType yotiprotoattr.ContentType, byteValue []byte) (interf
 
 	case yotiprotoattr.ContentType_JPEG,
 		yotiprotoattr.ContentType_PNG:
-		return byteValue, nil
+		return ParseImageValue(contentType, byteValue)
+
+	case yotiprotoattr.ContentType_UNDEFINED:
```
echarrod: I think it would be good to add a bit of logging here, I think we have that in the other SDKs.
Something like:
```go
log.Print("UNDEFINED Content type, attempting to parse as a string")
```

emmas-yoti: If we add the logging, then what's the difference between UNDEFINED and default?

echarrod: UNDEFINED has this comment in Attribute.proto:
> `UNDEFINED should not be seen, and is used as an error placeholder value.

So if something went wrong, it will be UNDEFINED, where as when it is a new type which is not recognized (yet), then it will be default

PR #91, opened 2020-01-02
SDK-1347 timeattribute flexible parsing

emmas-yoti:
### Add
 * Dependency on araddon/dateparser

### Change
 * TimeAttribute to support flexible parsing, allowing it to parse datetimes as well as dates.

### Comments
I'm not sure whether this is an approach we want to take with time attribute, as it doesn't contain information on what was parsed - since it leaves 'unused' attributes as the 0-value (or 1 for day) there's no way to tell the difference between 1st Jan and No Data.

### DIFF
--- a/attribute/date_attribute.go
+++ b/attribute/date_attribute.go
@@ -8,14 +8,14 @@ import (
 	"github.com/getyoti/yoti-go-sdk/v2/yotiprotoattr"
 )
 
-// TimeAttribute is a Yoti attribute which returns a time as its value
-type TimeAttribute struct {
+// DateAttribute is a Yoti attribute which returns a time as its value
+type DateAttribute struct {
 	attributeDetails
 	value *time.Time
 }
 
 // NewTime creates a new Time attribute
-func NewTime(a *yotiprotoattr.Attribute) (*TimeAttribute, error) {
+func NewDate(a *yotiprotoattr.Attribute) (*DateAttribute, error) {
 	parsedTime, err := time.Parse("2006-01-02", string(a.Value))
 	if err != nil {
 		log.Printf("Unable to parse time value of: %q. Error: %q", a.Value, err)
@@ -25,7 +25,7 @@ func NewTime(a *yotiprotoattr.Attribute) (*TimeAttribute, error) {
 
 	parsedAnchors := anchor.ParseAnchors(a.Anchors)
 
-	return &TimeAttribute{
+	return &DateAttribute{
 		attributeDetails: attributeDetails{
 			name:        a.Name,
 			contentType: a.ContentType.String(),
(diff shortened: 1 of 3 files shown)


echarrod:
Changeset - https://github.com/getyoti/yoti-go-sdk/compare/SDK-539-RenameGoFiles...SDK-1347-TimeattributeFlexibleParsing

Would be good to include a link to the changeset like this in the PR description if we're branching off another branch, so the reviewers don't have to look it up and work out which branch has been branched off 😃 👍 

RE: the `unused` attributes - do you mean if clients are initialising the struct manually, rather than using `NewTime`?
Or do you mean if they are using `NewTime`, but it's an invalid date?
If it's the latter - we're actually returning `nil` for the value, not the default `time.Time{}`, right?

emmas-yoti:
> RE: the `unused` attributes - do you mean if clients are initialising the struct manually, rather than using `NewTime`?
> Or do you mean if they are using `NewTime`, but it's an invalid date?

If they're using NewTime, and the attribute has the Value "2006-01-02", the resulting time.Time object will have the value corresponding to `Jan 1 2006 @ 00:00:00.000000000 UTC` which is the same value as if the attribute value is `2006-01-02T00:00:00.000000Z`

On the other side: if the attribute Value is the string `15:30:26` you get the time.Time object for `0-1-1T15:30:26.000000000Z` which is at least easy to filter for.

echarrod:
I think `Jan 1 2006 @ 00:00:00.000000000 UTC` makes sense in this instance because it's a _time_ attribute. In hindsight it might have better to call it a DateTimeAttribute, although I guess we are just reflecting what the Core Go dev team decided to call it (time.Time).

Hmm yes I don't think we want anyone to be passing in `15:30:26`, I'm thinking it might be good to put some validation on that, if it is allowed by the dateparser library.

attribute/date_attribute.go
```diff
@@ -8,14 +8,14 @@ import (
 	"github.com/getyoti/yoti-go-sdk/v2/yotiprotoattr"
 )
 
-// TimeAttribute is a Yoti attribute which returns a time as its value
-type TimeAttribute struct {
+// DateAttribute is a Yoti attribute which returns a time as its value
```
echarrod: Think we should update this comment too. Something like this:
```suggestion
// DateAttribute is a Yoti attribute which returns a date as *time.Time for its value
```

emmas-yoti:
Reply

> I think `Jan 1 2006 @ 00:00:00.000000000 UTC` makes sense in this instance because it's a _time_ attribute. In hindsight it might have better to call it a DateTimeAttribute, although I guess we are just reflecting what the Core Go dev team decided to call it (time.Time).
> 
> Hmm yes I don't think we want anyone to be passing in `15:30:26`, I'm thinking it might be good to put some validation on that, if it is allowed by the dateparser library.

Just updating. Since we've decided not to do flexible parsing - just to rename the Attribute, this will still validate against YYYY-MM-DD in NewDate, otherwise it'll return error.

PR #94, opened 2020-01-07
SDK-1359: Remove New() methods from Builders, handle nil values in Bu…

emmas-yoti:
…ild()

### DIFF
--- a/dynamic_sharing_service/dynamic_scenario_builder.go
+++ b/dynamic_sharing_service/dynamic_scenario_builder.go
@@ -41,7 +41,10 @@ func (builder *DynamicScenarioBuilder) Build() (DynamicScenario, error) {
 		builder.scenario.extensions = make([]interface{}, 0)
 	}
 	if builder.scenario.policy == nil {
-		policy, _ := (&DynamicPolicyBuilder{}).Build()
+		policy, err := (&DynamicPolicyBuilder{}).Build()
+		if err != nil {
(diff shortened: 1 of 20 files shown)


dynamic_sharing_service/dynamic_scenario_builder.go
```diff
@@ -45,6 +37,13 @@ func (builder *DynamicScenarioBuilder) WithCallbackEndpoint(endpoint string) *Dy
 
 // Build constructs the DynamicScenario
 func (builder *DynamicScenarioBuilder) Build() (DynamicScenario, error) {
+	if builder.scenario.extensions == nil {
+		builder.scenario.extensions = make([]interface{}, 0)
+	}
+	if builder.scenario.policy == nil {
+		policy, _ := (&DynamicPolicyBuilder{}).Build()
```
echarrod: We don't want to be ignoring the error value from Build() do we? 

i think gometalinter used to warn about this, i'm guessing golangci-lint doesn't

emmas-yoti: I wasn't sure, since this is for an empty Builder there should be no way it can fail to build. I can go either way with this though.

echarrod: There should be no way it can fail to build now, but always best to have a check in case the code changes later on! 👍

PR #95, opened 2020-01-08
SDK-1343: increase golang coverage

emmas-yoti:


attribute/document_details_attribute_test.go
```diff
@@ -24,6 +25,24 @@ func ExampleDocumentDetails_Parse() {
...
+	attribute, err := NewDocumentDetails(&proto)
+	if err != nil {
+		return
+	}
+	fmt.Printf(
+		"Document Type: %s, With %d Anchors",
+		attribute.Value().DocumentType,
+		len(attribute.Anchors()),
+	)
+	// Output: Document Type: PASSPORT, With 0 Anchors
```
echarrod: Is there a reason we didn't use asserts here? 
e.g. 
```go
	assert.Equal(t, "PASSPORT", attribute.Value().DocumentType)
	assert.Equal(t, 0, len(attribute.Anchors()))
```

I think it make it a bit clearer what we're actually testing

emmas-yoti: You can't use asserts in Examples.

Examples get compiled into the autgenerated documentation, so it's good to add them to Public methods.

echarrod: Ah ok, gotcha

requests/signed_message_test.go
```diff
@@ -121,3 +143,15 @@ func ExampleSignedRequest_WithPemFile() {
...
+func TestSignedRequest_WithPemFile_NotPemEncodedShouldError(t *testing.T) {
+	msg := SignedRequest{}.WithPemFile([]byte("not pem encoded"))
+	assert.ErrorContains(t, msg.Error, "Not PEM-encoded")
+}
+
+func TestSignedRequest_WithPemFile_NotRSAKeyShouldError(t *testing.T) {
+	msg := SignedRequest{}.WithPemFile([]byte(`-----BEGIN RSA PUBLIC KEY-----
+` + exampleKey + `
+-----END RSA PUBLIC KEY-----`))
+	assert.ErrorContains(t, msg.Error, "Not an ESA Private Key")
```
davidgrayston:
```suggestion
	assert.ErrorContains(t, msg.Error, "Not an RSA Private Key")
```

_(also need to change https://github.com/getyoti/yoti-go-sdk/blob/master/requests/signed_message.go#L96)_

PR #99, opened 2020-01-16
SDK-1027: Change anchor value to be a single string

echarrod:
We'll need to update the dev docs after this change too.

### DIFF
--- a/anchor/anchor_parser.go
+++ b/anchor/anchor_parser.go
@@ -26,30 +26,9 @@ var (
 func ParseAnchors(protoAnchors []*yotiprotoattr.Anchor) []*Anchor {
 	var processedAnchors []*Anchor
 	for _, protoAnchor := range protoAnchors {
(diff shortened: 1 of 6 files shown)


anchor/anchor_parser.go
```diff
@@ -45,11 +46,12 @@ func ParseAnchors(protoAnchors []*yotiprotoattr.Anchor) []*Anchor {
 					continue
 				}
 				anchorType = parsedAnchorType
-				extensions = append(extensions, value)
+				extension = value
+				break certs
```
emmas-yoti: We should... probably refactor this goto out, using an unexported function that returns.

echarrod: Fair point, I've updated in https://github.com/getyoti/yoti-go-sdk/pull/99/commits/226373528a1c1a69f018ffc6ead9eff60e9b5ba1

PR #104, opened 2020-02-11
SDK-1444: Remove Regex Validation and handle redacted aadhaar numbers

emmas-yoti:
### Changed
 * Removed regular expression validation from `DocumentDetails.Parse`


attribute/documentdetailsattribute.go
```diff
@@ -78,12 +73,12 @@ func NewDocumentDetails(a *yotiprotoattr.Attribute) (*DocumentDetailsAttribute,
 
 // Parse fills a DocumentDetails object from a raw string
 func (details *DocumentDetails) Parse(data string) (err error) {
-	if !documentDetailsValidateRegexp.MatchString(data) {
+	dataSlice := strings.Split(data, " ")
+
+	if len(dataSlice) < 3 {
 		return fmt.Errorf("Document Details data is invalid, %s", data)
 	}
```
MrBurtyyy: Can we add in a check to make sure that there are no empty strings in the split array?  I think this will also need tests which you can lift from the other SDK's

emmas-yoti: Have added this check just below.

.travis.yml
```diff
@@ -65,16 +48,16 @@ jobs:
...
-      go: "1.x"
+      go: "1.13.x"
       os: "osx"
     - <<: *test
-      go: "master"
+      go: "1.x"
       os: "linux"
     - <<: *test
-      go: "master"
+      go: "1.x"
```
echarrod: I think we should keep testing against `master` here too. We are allowing failures against it anyway, so we might as well keep it.

echarrod: Sorry, I now see it's been moved to the top of the list. I think it would be better to keep it at the bottom, both for the logical order, and for people viewing the changeset and thinking that it was removed (like I did!).

emmas-yoti: The reason it was moved is that the first one is tagged for later versions in the config file to reference it. Having the oldest version at the top of the file means touching more lines of code when/if we drop support for a version - however the master config should almost never change, so it makes more sense for that to be tagged as the reference configuration?

echarrod: I think the part where it tags it for later versions is:
```yml
  include:
    - &test
      stage: Test
      go: "master"
      os: "linux"
      before_script:
```
And then this declaration is below that, so we can keep the one above at the top, and move this one back to the bottom, right?

emmas-yoti: Have checked in a commit restoring master to bottom of list to show what I mean.

echarrod: As discussed, we'll leave it like this for now, and will reorder the versions to be most recent at top (master), to oldest at bottom in a separate changeset, so as not to interfere with the changes we've made here.

PR #108, opened 2020-02-19
Release 2.7.1

emmas-yoti:
### Changed
 * Document Details parsing to now handle redacted Aadhaar ID numbers (and other redacted ID numbers)

### Removed
 * Support for Go 1.10 and earlier. Go's release policy is to only support a version until there are two newer major releases. The current latest is 1.13, so official support extends only back to 1.12. Additionally, while Go promises backwards compatibility, Go 1.11 is the earliest version of Go with an official implementation of the Go modules feature which is necessary to fetch our dependencies. Due to upstream changes the previous workaround no longer functions.


.travis.yml
```diff
@@ -89,4 +73,19 @@ jobs:
         - $HOME/gopath/bin/goveralls -coverprofile=coverage.out -service=travis-ci
       if: type = pull_request OR branch = master
       after_success:
-        - coveralls
\ No newline at end of file
+        - coveralls
+
+    - stage: Analysis
+      name: Sonarcloud
+      if: type = pull_request
```
MrBurtyyy: Can we add `OR branch = master` here?

echarrod:
Would be good if we link to the release notes and explain a bit about dropping support for 1.10 and earlier in the 2nd bullet. Something like: 
> Dropped support for Go 1.10 and earlier. Go's [release policy](https://golang.org/doc/devel/release.html#policy) is to only support a version until there are two newer major releases. 1.13 is released now, and 1.14 will be [released soon](https://github.com/golang/go/milestones).

PR #116, opened 2020-03-27
SDK-1527: Sandbox Documentation

echarrod:
- Move .gitignore to examples folder, instead of having separate ones
- Added test_sandboxexample. Since the filename doesn't end in `_test`, it won't be run on every test execution (and then SetupSharingToken) would be called every time. We can leave this to being tested in the integration tests

@jlmartyres let me know if the new README.md looks good from an integrations perspective 👍


profile/sandbox/anchor.go
```diff
@@ -23,11 +27,22 @@ func VerifierAnchor(subtype string, timestamp time.Time, value string) Anchor {
...
+// It describes how an attribute has been provided to Yoti
+// (SOURCE Anchor) and how it has been verified (VERIFIER Anchor).
 type Anchor struct {
-	Type      string
-	Value     string
-	SubType   string
+	// Type of the Anchor - most likely either SOURCE or VERIFIER, but it's
+	// possible that new Anchor types will be added in future.
+	Type string
+	//	Value identifies the provider that either sourced or verified the attribute value.
```
davidgrayston, line 37:
```suggestion
	// Value identifies the provider that either sourced or verified the attribute value.
```

examples/profilesandbox/test_sandboxanchorexample.go
```diff
@@ -0,0 +1,59 @@
...
+
+	sandboxClient := sandbox.Client{
+		ClientSdkID: sandboxClientSdkId,
+		Key:         privateKey,
+	}
+
+	sourceAnchor := sandbox.SourceAnchor("NFC", time.Now().UTC(), "PASSPORT")
+	verifierAnchor := sandbox.VerifierAnchor("", time.Now().UTC(), "YOTI_ADMIN")
+
+	familyNameAttributeWithAnchors := sandbox.Attribute{}.
```
davidgrayston, line 34: Is is worth having an example providing anchors to pre-defined attributes like `WithGivenNames`, as it's likely integrators will start with these?

echarrod: Good point, updated in 8f75d50

profile/sandbox/anchor.go
echarrod: Corrected in 1fe608a

PR #117, opened 2020-04-01
SDK-1500: Use .env `YOTI_API_URL` for calls through YotiClient

echarrod:
Branched off https://github.com/getyoti/yoti-go-sdk/pull/116, see changes for this branch here:
https://github.com/getyoti/yoti-go-sdk/compare/SDK-1527-SandboxDocumentation...SDK-1500-UseEnvApiURLForYotiClient

We are already using this .env variable for Sandbox, this is just to update the YotiClient to do the same too. 

I opted to keep the `.OverrideAPIURL` in the sandbox example files (when we could instead just use the .env files), as I think it makes it clearer to the client that the client is not pointing at the same place

### TODO 
* [x] Merge https://github.com/getyoti/yoti-go-sdk/pull/116 into development

### DIFF
--- a/yoticlient.go
+++ b/yoticlient.go
@@ -8,6 +8,7 @@ import (
 	"io/ioutil"
 	"log"
 	"net/http"
+	"os"
 	"reflect"
 	"strconv"
 	"strings"
@@ -67,8 +68,7 @@ func (client *Client) doRequest(request *http.Request) (*http.Response, error) {
 	return client.httpClient.Do(request)
 }
 
-// OverrideAPIURL overrides the default API URL for this Yoti Client to permit
-// testing
+// OverrideAPIURL overrides the default API URL for this Yoti Client
 func (client *Client) OverrideAPIURL(apiURL string) {
 	client.apiURL = apiURL
 }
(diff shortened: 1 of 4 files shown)


yoticlient.go
```diff
@@ -88,10 +88,15 @@ func (client *Client) GetUserProfile(token string) (userProfile UserProfile, fir
...
 
 func (client *Client) getAPIURL() string {
-	if client.apiURL != "" {
-		return client.apiURL
+	if client.apiURL == "" {
+		if os.Getenv("YOTI_API_URL") != "" {
+			return os.Getenv("YOTI_API_URL")
+		} else {
+			return apiDefaultURL
+		}
```
MrBurtyyy, line 96:
```suggestion
		if value, ok := os.LookupEnv("YOTI_API_URL"); ok {
			return value
		}
		return apiDefaultURL
```

echarrod: Have updated in 31a425c, thought it was slightly clearer using `exists` instead of `ok`, and did a little restructuring to make the logic clearer, and also applied to the other place where we read this variable

PR #128, opened 2020-07-03
SDK-1495: Document images

echarrod:
- Reworked pre-commit to ensure it works on all OS
  - Use `go-unit-tests` from pre-commit-golang instead of custom script.
  - Use script for go-fmt to ensure cross-platform compatiblity.
  - Ignore deprecated package warning (for Protobuf package, will remove in major)

- Use `#!/usr/bin/env bash` as this can be picked up out of the box by windows too
- Fixed issue where extra `;` was being added to Base64URL, which browsers seem to handle fine, but thought it was worth fixing


gofmt.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 unset dirs files
 dirs=$(go list -f {{.Dir}} ./... | grep -v /yotiprotoshare/ | grep -v /yotiprotocom/ | grep -v /yotiprotoattr/)
     for d in $dirs
```
davidgrayston, line 4: not critical to fix, but the indentation in this file is inconsistent

echarrod: Fixed in 3f24f51

PR #137, opened 2020-07-15
SDK-1659: Split files out into sub-packages

echarrod:
- Move ensuring "address" attribute to address file, in attribute package. This was the piece of code which was making yoticlient have a high cyclomatic complexity, so...
- Reduced the limit of gocyclo to be 15 now (current highest is attribute.parser.go with 12), 15 is default recommended.
- Update remaining SDK references to v3
- Move sh files to sh folder, for some reason the permissions were reset after this, I think because I stashed and unstashed the files 😒 
- Refactor code into separate sub-packages:
  - web
  - shareurl
  - util
  - profile
  - aml

- Changed wording in README, it seems `go get` won't be useful for clients using modules, if they want the dependency in their SDK. Since modules are the chosen path for dependency management, I don't think it's pragmatic to mention `go get`, although people can still do this if they want
- Update go.mod to state `go 1.11` instead of `go 1.9`. We state 1.11+ on our README, and this also matched our Travis builds


README.md
```diff
@@ -63,20 +63,14 @@ Supported Go Versions:
...
 ## Installing the SDK
 
-_As of version **2.4.0**, [modules](https://github.com/golang/go/wiki/Modules) are used. This means it's not necessary to get a copy or fetch all dependencies as instructed below, as the Go toolchain will fetch them as necessary. You can simply add a `require github.com/getyoti/yoti-go-sdk/v2` to go.mod._
-
-To download and install the Yoti SDK and its dependencies, run the following command from your terminal:
-
-```Go
-go get "github.com/getyoti/yoti-go-sdk/v2"
-```
+_As of version **2.4.0**, [modules](https://github.com/golang/go/wiki/Modules) are used. This means it's not necessary to get a copy or fetch all dependencies as instructed below, as the Go toolchain will fetch them as necessary. You can simply add a `require github.com/getyoti/yoti-go-sdk/v3` to go.mod._
```
davidgrayston, line 66:
```suggestion
_As of version **2.4.0**, [modules](https://github.com/golang/go/wiki/Modules) are used. This means it's not necessary to get a copy or fetch all dependencies, as the Go toolchain will fetch them as necessary. You can simply add a `require github.com/getyoti/yoti-go-sdk/v3` to go.mod._
```

constants_for_test.go
```diff
@@ -4,5 +4,4 @@ const (
...
 	wrappedReceiptKey = "kyHPjq2+Y48cx+9yS/XzmW09jVUylSdhbP+3Q9Tc9p6bCEnyfa8vj38AIu744RzzE+Dc4qkSF21VfzQKtJVILfOXu5xRc7MYa5k3zWhjiesg/gsrv7J4wDyyBpHIJB8TWXnubYMbSYQJjlsfwyxE9kGe0YI08pRo2Tiht0bfR5Z/YrhAk4UBvjp84D+oyug/1mtGhKphA4vgPhQ9/y2wcInYxju7Q6yzOsXGaRUXR38Tn2YmY9OBgjxiTnhoYJFP1X9YJkHeWMW0vxF1RHxgIVrpf7oRzdY1nq28qzRg5+wC7cjRpS2i/CKUAo0oVG4pbpXsaFhaTewStVC7UFtA77JHb3EnF4HcSWMnK5FM7GGkL9MMXQenh11NZHKPWXpux0nLZ6/vwffXZfsiyTIcFL/NajGN8C/hnNBljoQ+B3fzWbjcq5ueUOPwARZ1y38W83UwMynzkud/iEdHLaZIu4qUCRkfSxJg7Dc+O9/BdiffkOn2GyFmNjVeq754DCUypxzMkjYxokedN84nK13OU4afVyC7t5DDxAK/MqAc69NCBRLqMi5f8BMeOZfMcSWPGC9a2Qu8VgG125TuZT4+wIykUhGyj3Bb2/fdPsxwuKFR+E0uqs0ZKvcv1tkNRRtKYBqTacgGK9Yoehg12cyLrITLdjU1fmIDn4/vrhztN5w="
-	attributeName     = "test_attribute_name"
```
davidgrayston, line 7: Could these live in a `yoti/test/constants.go` (package: test)?

web/endpoint.go
```diff
@@ -1,19 +1,21 @@
...
 
-func getAMLEndpoint(sdkID string) string {
+// GetAMLEndpoint gets the AML Endpoint URI
+func GetAMLEndpoint(sdkID string) string {
 	return fmt.Sprintf("/aml-check?appId=%s", sdkID)
 }
 
 // GetDynamicShareEndpoint gets the Dynamic Share Endpoint URI
-func getDynamicShareEndpoint(client clientInterface) (string, error) {
+func GetDynamicShareEndpoint(client ClientInterface) (string, error) {
```
davidgrayston, line 18: Perhaps we can move these into their own services/packages? (then we can keep them private within those packages)

yoti_client.go
```diff
@@ -121,13 +115,13 @@ func (client *Client) getActivityDetails(token string) (activity ActivityDetails
...
 
-	decryptedToken, err := decryptToken(token, client.Key)
+	decryptedToken, err := cryptoutil.DecryptToken(token, client.Key)
 
 	if err != nil {
 		return activity, errors.New("Unable to decrypt token")
 	}
 
-	endpoint := getProfileEndpoint(decryptedToken, client.GetSdkID())
+	endpoint := web.GetProfileEndpoint(decryptedToken, client.GetSdkID())
```
davidgrayston, line 124: Could we move this into the `profile` package - something like `profile/service.go`, which would handle the request?

The `yoti.Client` can just use these services (same for AML)

web/endpoint.go
echarrod: 👍 moved in db15f25, will make that unexported in the next commit

yoti_client.go
echarrod: Moved in db15f25, will make unexported in next commit

constants_for_test.go
echarrod: Moved in adc7bb8

.idea/.gitignore
```diff
@@ -0,0 +1,8 @@
+# Default ignored files
```
davidgrayston, line 1: Should we remove this folder?

requests/client.go
```diff
@@ -0,0 +1,28 @@
...
+
+	// DefaultHTTPErrorMessages maps HTTP error status codes to format strings
+	// to create useful error messages. -1 is used to specify a default message
+	// that can be used if an error code is not explicitly defined
+	DefaultHTTPErrorMessages = map[int]string{
+		-1: defaultUnknownErrorMessageConst,
+	}
+)
+
+//HttpClient is a mockable HTTP Client Interface
```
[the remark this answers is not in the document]
echarrod: Good spot, thought this would have been picked up by gofmt or goimports, fixed in dca433

.idea/.gitignore
echarrod: fixed in dca433

PR #138, opened 2020-07-15
SDK-1540: Doc Scan Sandbox

davidgrayston:
- [x] Merge #142 and #144 then rebase this PR

### Added
- Doc Scan sandbox support


_docs/PROFILE_SANDBOX.md
```diff
@@ -0,0 +1,30 @@
...
+import "github.com/getyoti/yoti-go-sdk/v3/profile/sandbox"
+```
+
+## Configuration
+The sandbox is initialised in the following way:
+```Go
+sandboxClient := sandbox.Client{
+		ClientSdkID: sandboxClientSdkId,
+		Key:         privateKey,
+	}
```
echarrod, line 19:
```suggestion
sandboxClient := sandbox.NewClient(sandboxClientSdkId, privateKey)
```

_docs/DOC_SCAN_SANDBOX.md
```diff
@@ -0,0 +1,25 @@
...
+You can reference the sandbox by adding the following import:
+
+```Go
+import "github.com/getyoti/yoti-go-sdk/v3/docscan/sandbox"
+```
+
+## Configuration
+The sandbox is initialised in the following way:
+```Go
+sandboxClient := sandbox.NewClient(sandboxClientSdkId, privateKey)
```
echarrod, line 16: I _think_ I made the profile Sandbox take `*rsa.PrivateKey` instead of []byte, because the bytes will be disposed once the key is read, and we'll need the key again when we initialise the regular Yoti Client. Will update this if that is the case, once I test it out

_examples/docscansandbox/demo_test.go
```diff
@@ -0,0 +1,338 @@
...
+
+	return wd
+}
+
+func newSandboxClient() *sandbox.Client {
+	key, err := ioutil.ReadFile(os.Getenv("YOTI_KEY_FILE_PATH"))
+	sdkID := os.Getenv("YOTI_SANDBOX_CLIENT_SDK_ID")
+	if err != nil {
+		panic(err)
+	}
```
echarrod, line 68: Might as well swap these lines around? If the code gets refactored, it makes it clear which error we're referring to
```suggestion
	if err != nil {
		panic(err)
	}
	sdkID := os.Getenv("YOTI_SANDBOX_CLIENT_SDK_ID")
```

_examples/docscansandbox/demo_test.go
```diff
@@ -0,0 +1,338 @@
...
+
+	configErr := c.client.ConfigureSessionResponse(c.getIFrameSessionID(), responseConfig)
+
+	if configErr != nil {
+		request, _ := json.Marshal(responseConfig)
+		return errors.New(string(request) + configErr.Error())
+	}
+
+	return configErr
```
echarrod, line 211: configErr will always be nil here, so might be clearer what we're returning by just returning nil here? 
```suggestion
	return nil
```

_examples/docscansandbox/demo_test.go
```diff
@@ -0,0 +1,338 @@
...
+	if !strings.Contains(source, text) {
+		return fmt.Errorf("Page source does not contain \"%s\"", text)
+	}
+	return nil
+}
+
+func (c *webContext) iSwitchToTheIframe() error {
+	iFrame, err := c.wd.FindElement(selenium.ByTagName, "iframe")
+	if err != nil {
+		panic(err)
```
echarrod, line 253: Looks like we're returning the error in most functions in this class, would it be better to return this error too instead of panicking?
```suggestion
		return err
```

_examples/docscansandbox/demo_test.go
```diff
@@ -0,0 +1,338 @@
...
+
+func (c *webContext) iWaitSeconds(seconds int) error {
+	time.Sleep(time.Duration(seconds) * time.Second)
+	return nil
+}
+
+func (c *webContext) elementContains(selector string, text string) error {
+	elem, err := c.wd.FindElement(selenium.ByCSSSelector, selector)
+	if err != nil {
+		panic(err)
```
echarrod, line 274: Same with the two in this function. Unless there's a reason we're panicking and not returning?

docscan/sandbox/request/check/document_text_data_check_test.go
```diff
@@ -0,0 +1,72 @@
...
+	"github.com/getyoti/yoti-go-sdk/v3/docscan/sandbox/request/filter"
+)
+
+func ExampleDocumentTextDataCheckBuilder() {
+	breakdown, err := report.NewBreakdownBuilder().
+		WithResult("some_result").
+		WithSubCheck("some_check").
+		Build()
+	if err != nil {
+		return
```
echarrod, line 17: Does returning here now make it quite hard to debug if there is an error? Would it be better to print the error? I imagine it will be displayed against the failing test
```suggestion
		fmt.Println(err.Error())
		return
```

_examples/docscansandbox/demo_test.go
davidgrayston: Yes, makes sense to return the error instead

_examples/docscansandbox/demo_test.go
davidgrayston: Makes sense to return error instead

docscan/sandbox/request/check/document_text_data_check_test.go
davidgrayston: I think I copied approach in other examples in the SDK - no harm in adding the print if it helps 👍

_docs/DOC_SCAN_SANDBOX.md
echarrod: Just looked at the sandbox profile example and we currently show:
```go
privateKey, err := cryptoutil.ParseRSAKey(pemFileBytes)
	assert.NilError(t, err)

	sandboxClient := sandbox.Client{
		ClientSdkID: sandboxClientSdkId,
		Key:         privateKey,
	}
```
For setting up the profile sandbox, and then the pemFileBytes are used to initialise the YotiClient, so all is OK.

Maybe we can introduce the NewClient to the profile Sandbox, and then clients have the option to specify the bytes as well.

echarrod: Created SDK-1677 for introducing NewClient to profile Sandbox too

PR #140, opened 2020-07-17
SDK-1660: Organise Examples

davidgrayston:
### Changed
- Moved `examples` into` _examples` to keep them separate from main package
- Each example now has its own `go.mod` file
- Profile sandbox example tests no longer need to be renamed to run


_examples/aml/main.go
```diff
@@ -7,6 +7,7 @@ import (
 	"strconv"
 
 	yoti "github.com/getyoti/yoti-go-sdk/v3"
+	aml "github.com/getyoti/yoti-go-sdk/v3/aml"
```
echarrod, line 10: We don't need these when the package is the same name (same as line above, think I'm removing these in https://github.com/getyoti/yoti-go-sdk/pull/139, so no need to change line above
```suggestion
	"github.com/getyoti/yoti-go-sdk/v3/aml"
```

davidgrayston: Rebased with your changes

_examples/aml/go.mod
```diff
@@ -1,4 +1,4 @@
-module github.com/getyoti/yoti-go-sdk/v3/examples
+module aml
```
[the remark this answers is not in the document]
davidgrayston: I think we can leave these as they are for now as they're not intended to be used by other packages

PR #142, opened 2020-07-20
SDK-1659: Response Errors

davidgrayston:
Following #137 

### Fixed
- Response body is now only appended when available (no trailing `:`)
- Reset the response body after reading so that the response can be used by callers

### Changed
- `defaultUnknownErrorMessageConst` is now the message string, rather than a format - status code is now always prefixed, e.g. `404: Profile not found`

### DIFF
--- a/dynamic/service_test.go
+++ b/dynamic/service_test.go
@@ -5,8 +5,10 @@ import (
 	"io/ioutil"
 	"net/http"
 	"strings"
+	"testing"
 
 	"github.com/getyoti/yoti-go-sdk/v3/test"
+	"gotest.tools/v3/assert"
 )
 
 type mockHTTPClient struct {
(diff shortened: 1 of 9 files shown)


dynamic/service_test.go
```diff
@@ -49,3 +51,60 @@ func ExampleCreateShareURL() {
...
+	tempError, temporary := err.(interface {
+		Temporary() bool
+	})
+	assert.Check(t, temporary && tempError.Temporary())
+}
+
+func TestCreateShareURL_Unsuccessful_404(t *testing.T) {
+	_, err := createShareUrlWithErrorResponse(404, "NOT FOUND")
+
+	assert.ErrorContains(t, err, "404: Application was not found: NOT FOUND")
```
echarrod, line 69: Are we set on this error layout? Or is this a recommended layout? I think it reads slightly better if we have:
```suggestion
	assert.ErrorContains(t, err, "404: NOT FOUND - Application was not found")
```
This has:
- Status code and reason together, which I think makes more logical sense
- A `-` instead of a 2nd `;`, I'm not a huge fan of having two `:` in one message

davidgrayston: I'll change the `:` to `-`, but in this case `NOT FOUND` is the HTTP response body, which is always appended

davidgrayston: @echarrod are you happy for me to change this?

From:
```
{CODE}: {MESSAGE}: {BODY}
```
to:
```
{CODE}: {MESSAGE} - {BODY}
```

davidgrayston: @echarrod I've made this change in f788eed - I've also update the test strings to make it clearer they are the response body

echarrod: Ok, yep looks good now

PR #145, opened 2020-07-31
SDK-1300: Refactor docscan builders

echarrod:
- Refactor Doc Scan builders:
  - use pointers in "With" methods and in corresponding structs
  - don't return incomplete objects if not correctly built

Doc Scan isn't released yet, so these won't constitute breaking changes


docscan/sandbox/request/response_config.go
```diff
@@ -34,14 +34,14 @@ func (b *ResponseConfigBuilder) WithCheckReports(checkReports CheckReports) *Res
...
-	responseConfig := ResponseConfig{
+func (b *ResponseConfigBuilder) Build() (*ResponseConfig, error) {
+	responseConfig := &ResponseConfig{
 		CheckReports: b.checkReports,
 		TaskResults:  b.taskResults,
 	}
 
 	if responseConfig.CheckReports == nil {
-		return responseConfig, errors.New("Check Reports must be provided")
+		return nil, errors.New("Check Reports must be provided")
```
lkramer, line 44: If this is error has the potential to reach the SDK user, I would make it a global variable.
```
var (
   ErrMissingCheckReporst = errors.New("Check Reports must be provided")
)
```
That way the SDK user a) has a list of potential errors so they know what they can expect, b) They can do comparison between the error they get and the global one for better error handling on their end c) if you want to change the wording of the error, you can do so without breaking client's assumption.

docscan/sandbox/request/check/report/breakdown.go
```diff
@@ -53,21 +53,21 @@ func (b *BreakdownBuilder) WithDetail(name string, value string) *BreakdownBuild
...
-func (b *BreakdownBuilder) Build() (Breakdown, error) {
-	breakdown := Breakdown{
+func (b *BreakdownBuilder) Build() (*Breakdown, error) {
+	breakdown := &Breakdown{
 		SubCheck: b.subCheck,
 		Result:   b.result,
 		Details:  b.details,
 	}
 
 	err := validate.NotEmpty(breakdown.SubCheck, "Sub Check cannot be empty")
```
lkramer, line 63: Not strictly in scope of this ticket, but I find this NotEmpty validator a bit pointless.

I think a better way of doing this would be:
```
if breakdown.SubCheck == nil {
     return nil, ErrEmptySubCheck
}
```
Where ErrEmptySubCheck is a predefined global variable.

docscan/sandbox/request/response_config.go
echarrod: This is an error thrown when clients are using our Sandbox SDK to build a request, which they then make to our sandbox service. So I imagine their only experience with this error might be that they encounter it once, fix it, and then never encounter it again. Additionally, this will only be encountered in their test code, they won't be calling our sandbox in production code.

As such, I don't think they'll be looking for this error and comparing it, so I'm not sure if it's worth making the suggestions, would you agree?

docscan/sandbox/request/check/report/breakdown.go
echarrod: I agree, seems much clearer to have it inline, and doesn't involve checking for the same error twice. 
Have removed in 456039f

echarrod: (Have kept error message inline based on reasoning above)

PR #146, opened 2020-07-31
SDK-1300: Create media package

echarrod:
Since we need a general "media" object for use in doc scan, and the current `Image` object we use is in profile/attribute/image (this is being moved already in the major v3.0.0 (unreleased), so moving it to the root `media` package won't cause any additional pain).
This PR is merging into development (v3.0.0) branch, which we can then merge into SDK-1300-docscan (which will be v3.1.0)

- Created media package, with media value, which our images now use


media/image.go
```diff
@@ -0,0 +1,22 @@
+package media
+
+import "fmt"
+
+const (
+	// ImageTypeJpeg JPEG format
+	ImageTypeJpeg string = "jpeg"
```
lkramer, line 7: Why not do it like this:

```
type ImageType string
const (
        // ImageTypeJpeg JPEG format
   	ImageTypeJpeg ImageType = "image/jpeg"

        // ImageTypePng PNG format
	ImageTypePng ImageType = "image/png"
)
```
It will make things a bit cleaner elsewhere in your code, since you a) now have the full mime string b) you have control over the type, so you don't need to worry about rogue strings.

echarrod: Good point. I was worried about supporting new image types if we added them, but we'd need to update the parsing code anyway, so might as well make this change. Done in 6d51d82

media/mediavalue.go
```diff
@@ -0,0 +1,21 @@
+package media
+
+import (
+	"encoding/base64"
+)
+
+// Image format of the image and the image data
+type Value struct {
```
lkramer, line 8: It seems to be identical to Media, is a different structure needed?

echarrod:
@lkramer & @davidgrayston, have reverted to embedding it the type in 93f95fe, as even though the initialiser needs to be as shown in the description, we can envelop this in a helper function and no other package needs to worry about that.

I also updated documentImages to be a slice of *media.Image in e5b4552, as this matches how we pass around other images, and should have been that way originally

echarrod:
I'll probably do an interactive rebase on the commits to fix them up, as there was a lot of back and forth, will do that after you've reviewed the above changes though, to make it clearer what's changed since you commented

media/image_test.go
```diff
@@ -0,0 +1,38 @@
...
+	result := NewImage(ImageTypePng, imageBytes)
+	expectedDataUrl := "data:image/png;base64," + imageBase64Value
+
+	assert.Equal(t, expectedDataUrl, result.Base64URL())
+}
+
+func TestImage_Base64Selfie_CreateJpegImage(t *testing.T) {
+	imageBytes := []byte("value")
+
+	result := NewImage(ImageTypeJpeg, imageBytes)
```
davidgrayston, line 25:
```suggestion
	result := NewJpegImage(imageBytes)
```

media/image_test.go
```diff
@@ -0,0 +1,38 @@
...
+}
+
+func TestImage_Base64Selfie_CreatePngImage(t *testing.T) {
+	imageBytes := []byte("value")
+
+	result := NewImage(ImageTypeJpeg, imageBytes)
+	expectedDataUrl := "data:image/jpeg;base64," + imageBase64Value
+
+	assert.Equal(t, expectedDataUrl, result.Base64URL())
+}
```
davidgrayston, line 38:
```suggestion
func TestImage_Base64Selfie_CreatePngImage(t *testing.T) {
	imageBytes := []byte("value")

	result := NewPngImage(imageBytes)
	expectedDataUrl := "data:image/png;base64," + imageBase64Value

	assert.Equal(t, expectedDataUrl, result.Base64URL())
}
```

media/image_test.go
```diff
@@ -0,0 +1,38 @@
...
+	"testing"
+
+	"gotest.tools/v3/assert"
+)
+
+const (
+	imageBase64Value = "dmFsdWU="
+)
+
+func TestImage_Base64Selfie_CreateImage(t *testing.T) {
```
davidgrayston, line 13:
```suggestion
func TestImage_Base64URL_CreateImage(t *testing.T) {
```

media/image_test.go
```diff
@@ -0,0 +1,38 @@
...
+func TestImage_Base64Selfie_CreateJpegImage(t *testing.T) {
+	imageBytes := []byte("value")
+
+	result := NewImage(ImageTypeJpeg, imageBytes)
+	expectedDataUrl := "data:image/jpeg;base64," + imageBase64Value
+
+	assert.Equal(t, expectedDataUrl, result.Base64URL())
+}
+
+func TestImage_Base64Selfie_CreatePngImage(t *testing.T) {
```
davidgrayston, line 31:
```suggestion
func TestImage_Base64URL_CreatePngImage(t *testing.T) {
```

media/image_test.go
```diff
@@ -0,0 +1,38 @@
...
+func TestImage_Base64Selfie_CreateImage(t *testing.T) {
+	imageBytes := []byte("value")
+
+	result := NewImage(ImageTypePng, imageBytes)
+	expectedDataUrl := "data:image/png;base64," + imageBase64Value
+
+	assert.Equal(t, expectedDataUrl, result.Base64URL())
+}
+
+func TestImage_Base64Selfie_CreateJpegImage(t *testing.T) {
```
davidgrayston, line 22:
```suggestion
func TestImage_Base64URL_CreateJpegImage(t *testing.T) {
```

media/image.go
```diff
@@ -0,0 +1,38 @@
...
+}
+
+// NewPngImage creates a new PNG image with the specified binary data
+func NewPngImage(data []byte) *Image {
+	return NewImage(ImageTypePng, data)
+}
+
+// Base64URL is the Image encoded as a base64 URL
+func (i *Image) Base64URL() string {
+	return base64URL(i.MIMEType, i.Data)
```
davidgrayston, line 37:
```suggestion
	return i.Value.Base64URL()
```

media/mediavalue_test.go
```diff
@@ -0,0 +1,35 @@
...
+
+import (
+	"encoding/base64"
+	"fmt"
+	"testing"
+
+	"gotest.tools/v3/assert"
+)
+
+func TestMedia_Base64Selfie_Png(t *testing.T) {
```
davidgrayston, line 11:
```suggestion
func TestMedia_Base64URL_Png(t *testing.T) {
```

media/mediavalue_test.go
```diff
@@ -0,0 +1,35 @@
...
+	imageBase64Value, media := createMedia("png")
+	expectedDataUrl := "data:image/png;base64," + imageBase64Value
+
+	assert.Equal(t, expectedDataUrl, media.Base64URL())
+}
+
+func TestMedia_Base64Selfie_Jpeg(t *testing.T) {
```
davidgrayston, line 18:
```suggestion
func TestMedia_Base64URL_Jpeg(t *testing.T) {
```

profile/attribute/image_attribute.go
```diff
@@ -30,7 +33,24 @@ func NewImage(a *yotiprotoattr.Attribute) (*ImageAttribute, error) {
...
+		imageType = media.ImageTypeJpeg
+
+	case yotiprotoattr.ContentType_PNG:
+		imageType = media.ImageTypePng
+
+	default:
+		return nil, errors.New("cannot create Image with unsupported type")
+	}
+
+	return media.NewImage(imageType, byteValue), nil
```
davidgrayston, line 55:
```suggestion
	switch contentType {
	case yotiprotoattr.ContentType_JPEG:
		return media.NewJpegImage(byteValue), nil

	case yotiprotoattr.ContentType_PNG:
		return media.NewPngImage(byteValue), nil

	default:
		return nil, errors.New("cannot create Image with unsupported type")
	}
```

media/image.go
echarrod: This just calls a method which then calls `base64URL(i.MIMEType, i.Data)`, but I guess it's better to go through that path

media/image_test.go
echarrod: Hmm thought I already changed that, might have got lost somewhere. Fixed in 29c66bc

PR #148, opened 2020-08-03
SDK-1661: Refactor errors

echarrod:
- Move `activity_details_errors` into `GetActivityDetails` function: we originally wanted to have all the possible errors from calling GetActivityDetails in this file. But now only one of these is used, and it's not possible to move all of the errors thrown in that class to the single errors file, since we now make use of the multi errors, as well as temporary errors. So in order to facilitate this we'd have to change to not be just simple constants, and have functions instead.
- don't capitalise start of error messages - see https://github.com/golang/go/wiki/CodeReviewComments#error-strings, this is included as "laundry list of common mistakes" (and flagged by Goland on code inspection)

We still have 4 error messages that I've left as being capitalised since they're referring to proper nouns, e.g.
```go
return "", fmt.Errorf("Sharing Profile not created (HTTP %d) %s", response.StatusCode, string(body))
```


echarrod:
> LGTM Like I said, I would personally have preferred the errors to be predefined, but it's more stylistic preference, so this is fine.

Since that wouldn't be a breaking change, if we can find a sensible way to have the dynamic errors in this central place, then we can try to do that in a future release 👍

PR #155, opened 2020-08-06
SDK-1300: An attempt to show an alternative way for handling media

lkramer:


profile/attribute/image_slice_attribute.go
```diff
@@ -42,18 +46,23 @@ func NewImageSlice(a *yotiprotoattr.Attribute) (*ImageSliceAttribute, error) {
...
-
-		result = append(result, imageValue)
+		switch i := item.Value.(type) {
+		case media.PNGImage:
+			result = append(result, i)
+		case media.JPEGImage:
+			result = append(result, i)
+		default:
+			return nil, fmt.Errorf("unexpected item type %T", i)
+		}
```
echarrod, line 59: Even if we don't go with this approach, I think it's definitely good to have this error handling here anyway 👍

profile/attribute/image_attribute.go
```diff
@@ -11,7 +11,7 @@ import (
 // ImageAttribute is a Yoti attribute which returns an image as its value
 type ImageAttribute struct {
 	attributeDetails
-	value *media.Image
+	value media.Media
```
echarrod, line 14: This approach does seem neater, **but** the thing I'm not keen on is having `ImageAttribute.value` and `ImageSliceAttribute.value` return media.Media, rather than media.Image. 

On the one hand, I think this stems from us calling them `Images` in every other SDK, and it's nice to be consistent.

On the other hand, is having `Images` needlessly adding complexity?

@davidgrayston what do you think?

profile/attribute/image_slice_attribute.go
```diff
@@ -11,7 +12,7 @@ import (
 // ImageSliceAttribute is a Yoti attribute which returns a slice of images as its value
 type ImageSliceAttribute struct {
 	attributeDetails
-	value []*media.Image
+	value []media.Media
```
echarrod, line 15: Do you think it's better to not use pointers with these images/media then?

lkramer: Media is an interface, you would not normally have pointers to interface

profile/attribute/image_attribute.go
lkramer: In all honesty, I would also like to refactor a bit around how attribute values are done, but I felt it was outside the scope of what I was trying to show here.

lkramer: The way I look at is that what you call Image is essentially one of two different things, a PNG or JPEG. I went with an approach with have these two as distinct types, both fulfilling an interface (media.Media). The validation is done when creating the ImageAttribute structure (and would be reason to not export the `value` variable so integrators can't override it with a non-checked value).

media/mediavalue.go
```diff
@@ -2,20 +2,15 @@ package media
...
+	"fmt"
 )
 
-// Value information about the media item, it's MIME type and binary data
-type Value struct {
-	MIMEType string
-	Data     []byte
+type Media interface {
+	Base64URL() string
+	MIMEType() MIMEType
```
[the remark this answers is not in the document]
lkramer: yes, I just pushed an updated which adds that and fixes the unit tests.

I also need to add comments here, I'll do that tomorrow.

lkramer:
> @echarrod in other SDKs we create `Media` for Doc Scan media responses - with this approach, we would need to create specific media based on the response content type, e.g. `application/pdf`.
> 
> To keep this simple, perhaps we should just have a `media.Media` struct that's used for everything?

That was my initial thought as well, however then you would have to other kinds of inspection to determine what kind of media it is. With this approach you can do
```
func DoStuff(m media.Media) {
  switch t := m.(type) {
     case PNGImage:
       // Do PNG stuff
     case JPEGImage:
      // Do JPEG stuff
     case ....
  } 
```
This is a more go-like way of doing it IMO.

You should be careful not to try to put a Square plug into a round hole. Other languages have features that may make some things easier (like generics, which Go doesn't have for now), but on the other hand Go have things that can make your life easier as well, and not taking advantage of them, makes your life unnecessarily difficult. In addition your integrators will not enjoy using your SDK if they have to treat it like a Java application.

profile/attribute/image_slice_attribute.go
echarrod: Ah yes, makes sense

echarrod:
The one reservation I have with this approach is that we want to be able to support new media types when we call `DocScanClient.GetMediaContent`, which I don't think this method will currently do. See Java for an example - https://github.com/getyoti/yoti-java-sdk/blob/6e4fe69d4cc02f237d3095152dd8ded4fcccbdea/yoti-sdk-impl/src/main/java/com/yoti/api/client/docs/DocScanService.java#L268

If we can allow for that somehow, then I'm happy to go with this

lkramer:
@echarrod @davidgrayston sorry for being a bit slow in understanding the requirements. I've pushed an update adding a `media.Generic` type.

PR #156, opened 2020-08-11
SDK-1300: Doc scan - polymorphic lists

echarrod:
We need a way of storing a polymorphic list: RequestedChecks, with the different sub types of RequestedChecks being allowed to be added to the list:
- RequestedLivenessCheck 
- RequestedFaceMatchCheck 
- RequestedDocumentAuthenticityCheck 

 As well as other potential future RequestedChecks implementations. 
This is solved in other SDKs with inheritance. Originally I tried embedded the base type in the child checks, but then they couldn't all be added to a single RequestedChecks list. 

The approach I've taken is to have a RequestedCheck**er** interface ([Go recommends adding `er` to the end of an interface](https://golang.org/doc/effective_go.html#interface-names)), which our checks then implement. And then anything that satisfies the interface can be added to the list. 

Once an approach is agreed here, we can do the same for RequestedTasks, and I think we'll need to do this when parsing the response too.


docscan/session/create/check/document_authenticity.go
```diff
@@ -1,28 +1,42 @@
...
+	checkType string
+	config    RequestedDocumentAuthenticityConfig
 }
 
-// RequestedDocumentAuthenticityConfig is the configuration applied when creating a Document Authenticity Check
-type RequestedDocumentAuthenticityConfig struct {
-	RequestedCheckConfig
+// Type is the type of the Requested Check
+func (c RequestedDocumentAuthenticityCheck) Type() string {
+	return c.checkType
```
davidgrayston, line 17:
```suggestion
// Type is the type of the Requested Check
func (c RequestedDocumentAuthenticityCheck) Type() string {
	return constants.IDDocumentAuthenticity
```

docscan/session/create/check/document_authenticity.go
```diff
@@ -1,28 +1,42 @@
...
+	"encoding/json"
+
 	"github.com/getyoti/yoti-go-sdk/v3/docscan/constants"
 )
 
 // RequestedDocumentAuthenticityCheck requests creation of a Document Authenticity Check
 type RequestedDocumentAuthenticityCheck struct {
-	RequestedCheck
-	Config RequestedDocumentAuthenticityConfig `json:"config"`
+	checkType string
```
davidgrayston, line 11:
```suggestion
```

docscan/session/create/check/document_authenticity.go
```diff
@@ -35,7 +49,10 @@ func NewRequestedDocumentAuthenticityCheckBuilder() *RequestedDocumentAuthentici
 	return &RequestedDocumentAuthenticityCheckBuilder{}
 }
 
-// Build creates a new DocumentAuthenticityCheck
+// Build builds the RequestedDocumentAuthenticityCheck
 func (b *RequestedDocumentAuthenticityCheckBuilder) Build() (*RequestedDocumentAuthenticityCheck, error) {
-	return NewRequestedDocumentAuthenticityCheck(b.config), nil
+	return &RequestedDocumentAuthenticityCheck{
+		checkType: constants.IDDocumentAuthenticity,
```
davidgrayston, line 55:
```suggestion
```

docscan/session/create/check/liveness.go
```diff
@@ -1,29 +1,44 @@
...
+		Type:   c.Type(),
+		Config: c.Config(),
+	})
 }
 
 // RequestedLivenessConfig is the configuration applied when creating a Liveness Check
 type RequestedLivenessConfig struct {
-	RequestedCheckConfig
-	MaxRetries   *int   `json:"max_retries"`
+	// RequestedCheckConfig
```
lkramer, line 40: Did you mean to remove this?

docscan/session/create/check/document_authenticity.go
echarrod: Ah yes I did this so we didn't add the constant twice, but removing it from the build method, and having this line as you suggested would make more sense I think

docscan/session/create/check/liveness.go
```diff
@@ -36,34 +51,34 @@ func NewRequestedLivenessCheckBuilder() *RequestedLivenessCheckBuilder {
 type RequestedLivenessCheckBuilder struct {
```
lkramer, line 51: I am not a huge fan of this builder structure. I'm not sure why purpose it serves.

Why not simply do:
```
func NewRequestedLivenessCheck(c *RequestedLivenessConfig) *RequestedLivenessCheck {
   return &RequestedLivenessCheck{
		checkType: constants.Liveness,
		config:    config,
	}
}
```

docscan/session/create/check/document_authenticity.go
echarrod: Fixed in 11cfa6c

docscan/session/create/check/liveness.go
echarrod: We've gone with builders for all of the Doc Scan payloads so that we could add new items to the payload, e.g. we would have a second argument in the function you specify:
```go
func NewRequestedLivenessCheck(c *RequestedLivenessConfig, priority int) *RequestedLivenessCheck {
   return &RequestedLivenessCheck{
		checkType: constants.Liveness,
		config:    config,
                priority:   priority
	}
}
```

Not a great example, as it could go in config, but the principle of adding/removing items from the constructor is hidden from the user (if they use it as intended), if they use the builder. And so we can make & break, without worrying about backwards compatibility

docscan/session/create/check/liveness.go
echarrod: Yes I meant to remove, have done that in d70a6d2

docscan/session/create/check/liveness.go
lkramer: yeah, I would think it would be preferable to hold those things in the Config. 

as for default values, you could still have that:
```
func NewRequestedLivenessCheck(c *RequestedLivenessConfig) *RequestedLivenessCheck {
   priority := c.Priority
   if priority == PriorityUndefined {
      priority = PriorityDefault
   }
   return &RequestedLivenessCheck{
		checkType: constants.Liveness,
		config:    config,
                priority: priority,
	}
}
```

docscan/session/create/check/liveness.go
```diff
@@ -36,34 +49,33 @@ func NewRequestedLivenessCheckBuilder() *RequestedLivenessCheckBuilder {
...
 	livenessType string
 	maxRetries   *int
-	//TODO: do we want this with pointer so we can set to nil?
 }
 
 // ForZoomLiveness sets the liveness type to "ZOOM"
-func (builder *RequestedLivenessCheckBuilder) ForZoomLiveness() *RequestedLivenessCheckBuilder {
-	return builder.ForLivenessType(constants.Zoom)
+func (b *RequestedLivenessCheckBuilder) ForZoomLiveness() *RequestedLivenessCheckBuilder {
+	return b.ForLivenessType(constants.Zoom)
```
lkramer, line 56: I noticed you declared a separate `constants` package. In Go, normally constants/enums should be declared close to where they are relevant.

It's also worth considering using type aliases:
```
type LivenessType string
const (
   Zoom LivenessType = "ZOOM"
)
```
And then make the value in the struct be of that type to prevent some rogue person accidently assigning "ID_DOCUMENT" to liveness.

docscan/session/create/check/liveness.go
lkramer: So looking a bit closer, I think maybe the issue here is that `RequestedLivenessConfig` is supposed to marshal directly into the config in the structure send over the wire. Essentially your builder struct is the actual config from the integraters POV.

This is probably OK actually, but I would drop the setters, and just keep the values exported so the integrator can do:
```
   builder := &RequestedLivenessCheckBuilder{
     LivenessType: ZoomLivenessType,
     MaxRetries: 5,
  }
```

docscan/session/create/check/liveness.go
```diff
@@ -36,34 +49,33 @@ func NewRequestedLivenessCheckBuilder() *RequestedLivenessCheckBuilder {
 type RequestedLivenessCheckBuilder struct {
 	livenessType string
 	maxRetries   *int
```
lkramer, line 51: Why is this a pointer?

echarrod: We want the default value to be nil, not 0

docscan/session/create/check/requested_check.go
```diff
@@ -1,11 +1,12 @@
 package check
 
-// RequestedCheck requests creation of a Check to be performed on a document
-type RequestedCheck struct {
-	Type   string               `json:"type"`
-	Config RequestedCheckConfig `json:"config"`
+// RequestedChecker requests creation of a Check to be performed on a document
+type RequestedChecker interface {
```
davidgrayston, line 4: I don't think this needs the `-er` suffix as it's not a one-method interface

lkramer: I would actually take this one step further and say it doesn't add value, since it's empty. It would be fine to just use `interface{}`

docscan/session/create/check/liveness.go
lkramer: I think I'm withdrawing my objections here. I have some minor concerns around int as a pointer and a suggestion to type alias the const values, but we're addressing that elsewhere.

docscan/session/create/check/requested_check.go
```diff
@@ -1,11 +1,12 @@
...
+	Type() string
+	Config() RequestedCheckConfig
+	MarshalJSON() ([]byte, error)
 }
 
 // RequestedCheckConfig is the configuration applied when creating a Check
-type RequestedCheckConfig struct {
+type RequestedCheckConfig interface {
```
lkramer, line 11: I suggest leaving this out, unless you plan on adding some methods. It's not different from `interface{}`

docscan/session/create/check/liveness.go
echarrod: I've made this change in 1574508, I've left them unexported if they don't need to be, and we can always make them exported at some point before releasing if need be.

docscan/session/create/check/requested_check.go
echarrod: Good catch, reverted in 0e9669d

docscan/session/create/check/requested_check.go
echarrod: I think i did this so that we can add methods to this interface later down the line, without it being a breaking change. 

Also with methods like:
```go
// Config is the configuration of the Requested Check
func (c RequestedDocumentAuthenticityCheck) Config() RequestedCheckConfig {
	return RequestedCheckConfig(
		c.config,
	)
}
```
it's a bit more clear what we expect, rather than:
```go
// Config is the configuration of the Requested Check
func (c RequestedDocumentAuthenticityCheck) Config() interface{} {
	return interface{}(
		c.config,
	)
}
```

docscan/session/create/check/liveness.go
echarrod: Have changed to not be a pointer in 465ba75, with a test to check it's omitted from the payload if not specified

docscan/session/create/check/requested_check.go
lkramer: Yeah, OK. Fair enough

docscan/session/create/sdk_config_test.go
```diff
@@ -0,0 +1,29 @@
...
+		WithFontColour("#ff0000").
+		WithLocale("fr_FR").
+		WithPresetIssuingCountry("USA").
+		WithPrimaryColour("#aa1111").
+		WithSecondaryColour("#bb2222").
+		WithSuccessUrl("https://successurl.com").
+		Build()
+
+	if err != nil {
+		fmt.Fprintf(os.Stdout, "error: %s", err.Error())
```
lkramer, line 22: Why not just `fmt.Printf(...) ` ?

docscan/session/create/sdk_config_test.go
```diff
@@ -0,0 +1,29 @@
...
+		return
+	}
+
+	data, _ := json.Marshal(notifications)
```
lkramer, line 26: So I know this is an example, but since you're checking errors elsewhere it seems inconsistent not to do it here.

If you want to get out of it easy a standard way of doing error handling in examples is to do this:
```
if err != nil {
   panic(err)
}
```

docscan/session/create/sdk_config_test.go
```diff
@@ -0,0 +1,29 @@
+package create
+
+import (
+	"encoding/json"
+	"fmt"
+	"os"
+)
+
+func ExampleSdkConfigBuilder_Build() {
```
lkramer, line 9: Is this an example or a test?

docscan/session/create/sdk_config_test.go
echarrod: My thinking process was that if there were any errors, then we'd find out on the line below when we write the result. Do you think it would give us more information if we checked this error too?

docscan/session/create/sdk_config_test.go
echarrod: I was going to add tests later on, this was just the intial PR to get feedback on the use of interfaces for checks. This stuff will be reviewed at least 3 more times after this: from Doc-Scan client -> Doc-Scan, Doc-Scan -> development, and in the release branch.

docscan/session/create/sdk_config_test.go
echarrod: Ah, I thought Fprintf was needed to give a good comparison when tests failed, but just tried with Printf and all looks good! Have updated that instance, and will update other instances in a separate PR

echarrod: Fixed in e16625a

docscan/session/create/check/liveness.go
```diff
@@ -34,36 +45,35 @@ func NewRequestedLivenessCheckBuilder() *RequestedLivenessCheckBuilder {
...
-	return builder.ForLivenessType(constants.Zoom)
+func (b *RequestedLivenessCheckBuilder) ForZoomLiveness() *RequestedLivenessCheckBuilder {
+	return b.ForLivenessType(zoom)
 }
 
 // ForLivenessType sets the liveness type on the builder
-func (builder *RequestedLivenessCheckBuilder) ForLivenessType(livenessType string) *RequestedLivenessCheckBuilder {
-	builder.livenessType = livenessType
-	return builder
+func (b *RequestedLivenessCheckBuilder) ForLivenessType(livenessType livenessType) *RequestedLivenessCheckBuilder {
```
lkramer, line 58: You sure you want this to be exported? There seems to be just one option.

If you do want it exported, you should make `livenessType` exported as well.

docscan/session/create/filter/constants.go
```diff
@@ -0,0 +1,19 @@
+package filter
+
+type documentType string
+
+const identity documentType = "ID_DOCUMENT"
+
+// type restrictions string
```
lkramer, line 7: Is this supposed to be removed?

docscan/session/create/sdk_config_test.go
lkramer: OK, I would put a big fat TODO on it then, it least it's clear something needs to be done before going to prod (not usually fan of TODOs)

docscan/session/create/sdk_config_test.go
lkramer: It's more about habit. Some rough examples on stack overflow etc will gloss over error handling, but I would never want production code not to have it, and if we have examples where we skip error handling it might give the wrong impression to people using the SDK.

echarrod: Yes I agree it sets a good example, I'll make a fix to update it everywhere in a separate PR

docscan/session/create/sdk_config_test.go
echarrod: I am a fan of TODOs in code, but only if they're all removed before release! 😁 which obviously has the potential to sometimes be missed...
Will add

docscan/session/create/check/liveness.go
echarrod: These were added with strings originally so that the SDKs could take advantage of new liveness types without an update. Just realised if we make this an aliased type, it is no longer possible to do that. This is the case for all the other builders too.

I think I might return these to strings, but we can still keep them in the same places to aid separation of concerns. 
@davidgrayston would you agree that I'll need to keep these as strings?

docscan/session/create/filter/constants.go
echarrod: They are constants for Document Restrictions, which I have started adding, but don't make use of these constants yet. Will uncomment and use for the next merge into Doc-Scan

docscan/session/create/check/liveness.go
davidgrayston: Other SDKs allow any string to be provided, so I think it would make sense to do the same here. e.g. https://github.com/getyoti/yoti-dotnet-sdk/blob/master/src/Yoti.Auth/DocScan/Session/Create/Check/RequestedLivenessCheckBuilder.cs#L15

docscan/session/create/sdk_config_test.go
echarrod: Based on https://blog.golang.org/examples, should examples have _test on the end as well?

echarrod: > As with typical tests, examples are functions that reside in a package's _test.go files. Unlike normal test functions, though, example functions take no arguments and begin with the word Example instead of Test.

docscan/session/create/check/liveness.go
echarrod: Yep, reverted back in 9c32f03

echarrod: > You sure you want this to be exported? There seems to be just one option.
> 
> If you do want it exported, you should make `livenessType` exported as well.

Sorry, never answered this - even though there's only one option now, we want to allow for future options to be added in the future, with the SDK being able to take advantage of it before we explicitly add a supported method

docscan/session/create/check/constants.go
```diff
@@ -0,0 +1,6 @@
+package check
+
+const (
+	liveness = "LIVENESS"
```
davidgrayston, line 4: Should `liveness = "LIVENESS"` live in `docscan/constants/constants.go` with the other check types?
(or should the other check type constants move here?)

docscan/session/create/filter/constants.go
```diff
@@ -0,0 +1,12 @@
+package filter
+
+const (
+	includeList string = "WHITELIST"
+	excludeList string = "BLACKLIST"
+
+	identityDocument string = "ID_DOCUMENT"
+
+// 	orthogonalRestriction string = "ORTHOGONAL_RESTRICTIONS"
+// 	documentRestriction   string = "DOCUMENT_RESTRICTIONS"
```
davidgrayston, line 10: As discussed, these will be used in the parent branch

docscan/session/create/sdk_config_test.go
```diff
@@ -0,0 +1,28 @@
...
+		Build()
+
+	if err != nil {
+		fmt.Printf("error: %s", err.Error())
+		return
+	}
+
+	data, _ := json.Marshal(notifications)
+	fmt.Println(string(data))
+	// Output: {"allowed_capture_methods":"CAMERA","primary_colour":"#aa1111","secondary_colour":"#bb2222","font_colour":"#ff0000","locale":"fr_FR","preset_issuing_country":"USA","success_url":"https://successurl.com"}
```
davidgrayston, line 27: `error_url` should be present in this JSON - `WithErrorUrl()` is currently setting the success URL

docscan/session/create/check/constants.go
echarrod: Will reorganise the constants before release, after all the Doc Scan stuff has been added 👍

docscan/session/create/sdk_config_test.go
echarrod: Good catch, fixed in fc4fef1

docscan/session/create/sdk_config_test.go
lkramer: oh! I did not know that. Looks fine then.

PR #166, opened 2020-08-13
SDK-1300: Doc Scan Client - Create

echarrod:
These changes contain the "Create" part of the Doc Scan service, along with some other (not extensively tested yet) parts. 

Have created the PR before working on the "Retrieve" part, so the work doesn't get too large. 

Some of the changes were already reviewed in https://github.com/getyoti/yoti-go-sdk/pull/156

Will work on improving the coverage later on, just trying to get this merged so the integration tests can be worked earlier.


docscan/client.go
```diff
@@ -0,0 +1,94 @@
...
+func (client *Client) getHTTPClient() requests.HttpClient {
+	if client.HTTPClient != nil {
+		return client.HTTPClient
+	}
+	return http.DefaultClient
+}
+
+// CreateSession creates a Doc Scan session using the supplied session specification
+func (client *Client) CreateSession(sessionSpec *create.SessionSpecification) (*create.SessionResult, error) {
+	return service.CreateSession(client.getHTTPClient(), client.SdkID, client.Key, client.getAPIURL(), sessionSpec, nil)
```
davidgrayston, line 68: It might be neater to inject `client.getHTTPClient()`, `client.SdkID`, `client.Key` and `client.getAPIURL()` into the service, then call with the parameters related to this method: `service.CreateSession(sessionSpec)`

This might not be easy to do neatly this as we have to consider `OverrideAPIURL` and ensure this is propagated to the service.

I think it might make sense to drop `docscan/service/session.go` and move the request builder code into the client

docscan/service/session.go
```diff
@@ -0,0 +1,204 @@
...
+		return nil, err
+	}
+
+	var responseBytes []byte
+	responseBytes, err = ioutil.ReadAll(response.Body)
+	if err != nil {
+		return nil, err
+	}
+
+	contentType := response.Header.Get("Content-type") // TODO: check this
```
davidgrayston, line 134: I think we should be defensive here and account for paramters, e.g. `Content-Type: text/html; charset=UTF-8` (split on `;` and take the first part)

Other languages/libraries will return this as an array, where we would take the first part.

docscan/service/session_test.go
```diff
@@ -0,0 +1,57 @@
...
+
+func TestCreateSession(t *testing.T) {
+	key, _ := rsa.GenerateKey(rand.Reader, 1024)
+
+	var clientSessionTokenTTL int = 100
+	var clientSessionToken string = "8c91671a-7194-4ad7-8483-32703b965cfc"
+	var sessionId string = "c87c4f2a-13fd-4cc8-a0e4-f1637cf32f71"
+
+	jsonResponse := fmt.Sprintf(`{"client_session_token_ttl":%d,"client_session_token":"%s","session_id":"%s"}`, clientSessionTokenTTL, clientSessionToken, sessionId)
+	// jsonResponse := fmt.Sprintf(, clientSessionTokenTTL, clientSessionToken, sessionId)
```
davidgrayston, line 39:
```suggestion
```

docscan/service/session_test.go
```diff
@@ -0,0 +1,57 @@
...
+}
+
+func TestCreateSession(t *testing.T) {
+	key, _ := rsa.GenerateKey(rand.Reader, 1024)
+
+	var clientSessionTokenTTL int = 100
+	var clientSessionToken string = "8c91671a-7194-4ad7-8483-32703b965cfc"
+	var sessionId string = "c87c4f2a-13fd-4cc8-a0e4-f1637cf32f71"
+
+	jsonResponse := fmt.Sprintf(`{"client_session_token_ttl":%d,"client_session_token":"%s","session_id":"%s"}`, clientSessionTokenTTL, clientSessionToken, sessionId)
```
davidgrayston, line 38: You could create this string using `create.SessionResult{}`?

It's good to keep the test data separate, so it's fine like this

docscan/session/create/check/document_authenticity_test.go
```diff
@@ -0,0 +1,21 @@
...
+	if err != nil {
+		fmt.Printf("error: %s", err.Error())
+		return
+	}
+
+	data, _ := json.Marshal(check)
+	fmt.Println(string(data))
+	// Output: {"type":"ID_DOCUMENT_AUTHENTICITY","config":{}}
+
+	// TODO: check we want config={}?
```
davidgrayston, line 20: I think we can remove this comment - we confirmed that omitting was more desirable, but empty `{}` was also fine

docscan/session/create/filter/document_restrictions.go
```diff
@@ -0,0 +1,49 @@
...
+	}
+}
+
+// WithCountryCode adds a country code to the Restriction
+func (b *RequestedDocumentRestrictionBuilder) WithCountryCode(countryCode string) *RequestedDocumentRestrictionBuilder {
+	b.countryCodes = append(b.countryCodes, countryCode)
+	return b
+}
+
+// TODO: need country code? Looks like Python just has plural
```
davidgrayston, line 29: We can remove the singular methods (other SDKs don't have this)

[Java has `withCountries` and `withDocumentTypes`](https://github.com/getyoti/yoti-java-sdk/blob/master/yoti-sdk-impl/src/main/java/com/yoti/api/client/docs/session/create/filters/SimpleDocumentRestrictionBuilder.java)

docscan/session/create/filter/constants.go
```diff
@@ -0,0 +1,12 @@
+package filter
+
+const (
+	includeList string = "WHITELIST"
+	excludeList string = "BLACKLIST"
+
+	identityDocument string = "ID_DOCUMENT"
+
+// 	orthogonalRestriction string = "ORTHOGONAL_RESTRICTIONS"
+// 	documentRestriction   string = "DOCUMENT_RESTRICTIONS"
```
davidgrayston, line 10: Will these filters be implemented on a separate PR?

docscan/session/create/filter/orthogonal_restrictions_filter.go
```diff
@@ -0,0 +1,65 @@
...
+// NewRequestedOrthogonalRestrictionsFilterBuilder creates a new RequestedOrthogonalRestrictionsFilterBuilder
+func NewRequestedOrthogonalRestrictionsFilterBuilder() *RequestedOrthogonalRestrictionsFilterBuilder {
+	return &RequestedOrthogonalRestrictionsFilterBuilder{
+		countryRestriction: &CountryRestriction{},
+		typeRestriction:    &TypeRestriction{},
+	}
+}
+
+// WithIncludeCountryList sets an "INCLUDE" list of country codes on the filter
+func (b *RequestedOrthogonalRestrictionsFilterBuilder) WithIncludeCountryList(countryCodes []string) *RequestedOrthogonalRestrictionsFilterBuilder {
```
davidgrayston, line 24: I think one of these might read better

```suggestion
func (b *RequestedOrthogonalRestrictionsFilterBuilder) WithIncludedCountries(countryCodes []string) *RequestedOrthogonalRestrictionsFilterBuilder {
```

```suggestion
func (b *RequestedOrthogonalRestrictionsFilterBuilder) WithIncludedCountryCodes(countryCodes []string) *RequestedOrthogonalRestrictionsFilterBuilder {
```

docscan/session/create/filter/orthogonal_restrictions_filter.go
```diff
@@ -0,0 +1,65 @@
...
+	b.countryRestriction = &CountryRestriction{
+		includeList,
+		countryCodes,
+	}
+	return b
+}
+
+// WithExcludeCountryList sets an "EXCLUDE" list of country codes on the filter
+func (b *RequestedOrthogonalRestrictionsFilterBuilder) WithExcludeCountryList(countryCodes []string) *RequestedOrthogonalRestrictionsFilterBuilder {
```
davidgrayston, line 33:
```suggestion
func (b *RequestedOrthogonalRestrictionsFilterBuilder) WithExcludedCountryCodes(countryCodes []string) *RequestedOrthogonalRestrictionsFilterBuilder {
```

docscan/session/create/filter/orthogonal_restrictions_filter.go
```diff
@@ -0,0 +1,65 @@
...
+	b.countryRestriction = &CountryRestriction{
+		excludeList,
+		countryCodes,
+	}
+	return b
+}
+
+// WithIncludeDocumentTypeList sets an "INCLUDE" list of document types on the filter
+func (b *RequestedOrthogonalRestrictionsFilterBuilder) WithIncludeDocumentTypeList(documentTypes []string) *RequestedOrthogonalRestrictionsFilterBuilder {
```
davidgrayston, line 42:
```suggestion
func (b *RequestedOrthogonalRestrictionsFilterBuilder) WithIncludedDocumentTypes(documentTypes []string) *RequestedOrthogonalRestrictionsFilterBuilder {
```

docscan/session/create/filter/orthogonal_restrictions_filter.go
```diff
@@ -0,0 +1,65 @@
...
+	b.typeRestriction = &TypeRestriction{
+		includeList,
+		documentTypes,
+	}
+	return b
+}
+
+// WithExcludeDocumentTypeList sets an "EXCLUDE" list of document types on the filter
+func (b *RequestedOrthogonalRestrictionsFilterBuilder) WithExcludeDocumentTypeList(documentTypes []string) *RequestedOrthogonalRestrictionsFilterBuilder {
```
davidgrayston, line 51:
```suggestion
func (b *RequestedOrthogonalRestrictionsFilterBuilder) WithExcludedDocumentTypes(documentTypes []string) *RequestedOrthogonalRestrictionsFilterBuilder {
```

docscan/session/create/filter/document_filter.go
```diff
@@ -0,0 +1,47 @@
+package filter
+
+// RequestedDocumentFilter filters for a required document, allowing specification of restrictive parameters
+type RequestedDocumentFilter struct {
+	DocumentTypes []string `json:"document_types"`
+	CountryCodes  []string `json:"country_codes"`
+}
```
davidgrayston, line 7: This should only have `json:"type"`, and be embedded in `RequestedDocumentRestrictionsFilter` and `RequestedOrthogonalRestrictionsFilter`

docscan/client.go
echarrod: Moved into the client in 4e7b2cc. Partly to match Doc Scan Sandbox, and also to mitigate the chance of breaking things if we change something which would now be internal do docscan.client

docscan/service/session_test.go
echarrod: Do you mean create it with a struct initializer and then serialize it?

echarrod:
@davidgrayston have moved the Document Filters work to https://github.com/getyoti/yoti-go-sdk/tree/SDK-1459-DocumentFilters, will merge those in a separate PR

docscan/service/session_test.go
davidgrayston: Yes, but I think it's fine as you have it

PR #167, opened 2020-08-14
SDK-1300: Add json subtype parsing with custom JSON unmarshalling

echarrod:
This is a proposed solution for dealing with the handling of various subtypes when parsing the GetSession response.

CheckResponses can be (based on the value of the `type` field):
- CheckResponse
  - AuthenticityCheckResponse
  - FaceMatchCheckResponse
  - LivenessCheckResponse
  - TextDataCheckResponse
  - may be others in future

Likewise, a GeneratedChecks can be:
- GeneratedCheck 
  - GeneratedTextDataCheckResponse
  - may be others in future

So we want to add convenience methods to _filter_:
- `GetSessionResult.Checks` to return the the above subsets listed, and
- `TaskResponse.GeneratedChecks` to only return the `GeneratedTextDataCheckResponses`

There are two other "subtypes" which will need filtering too, but thought I'd check this was the right approach before applying to those as well.

At first I had:
```go
type TaskResponse struct {
	GeneratedChecks         []*GeneratedCheckResponse `json:"generated_checks"`
	GeneratedTextDataChecks []*GeneratedTextDataCheckResponse
}
```
instead of what I currently have:
```go
type TaskResponse struct {
	GeneratedChecks         []*GeneratedCheckResponse `json:"generated_checks"`
	generatedTextDataChecks []*GeneratedTextDataCheckResponse
}

// GeneratedTextDataChecks  filters the checks, returning only text data checks
func (t *TaskResponse) GeneratedTextDataChecks() []*GeneratedTextDataCheckResponse {
	return t.generatedTextDataChecks
}
```

But I thought this 2nd approach makes it a bit clearer that they aren't two distinct slices, and `GeneratedChecks` also includes the items in `generatedTextDataChecks`

Another approach could be to have `GetSessionResult.Checks` and `TaskResponse.GeneratedChecks` return a slice of interfaces, with the subtypes each implementing the values on the interface, but I thought the approach in this PR was neater.

### DIFF
--- a/docscan/client.go
+++ b/docscan/client.go
@@ -153,7 +153,7 @@ func (c *Client) DeleteSession(sessionID string) error {
 }
 
 // GetMediaContent retrieves media related to a Yoti Doc Scan session based on the supplied media ID
-func (c *Client) GetMediaContent(sessionID, mediaID string) (*media.Media, error) { // TODO: change to media.Value
+func (c *Client) GetMediaContent(sessionID, mediaID string) (*media.Media, error) {
 	request, err := (&requests.SignedRequest{
 		Key:        c.Key,
 		HTTPMethod: http.MethodGet,
--- a/docscan/client.go
+++ b/docscan/client.go
@@ -153,7 +153,7 @@ func (c *Client) DeleteSession(sessionID string) error {
 }
 
 // GetMediaContent retrieves media related to a Yoti Doc Scan session based on the supplied media ID
-func (c *Client) GetMediaContent(sessionID, mediaID string) (*media.Media, error) {
+func (c *Client) GetMediaContent(sessionID, mediaID string) (media.Media, error) {
 	request, err := (&requests.SignedRequest{
 		Key:        c.Key,
 		HTTPMethod: http.MethodGet,
@@ -183,7 +183,8 @@ func (c *Client) GetMediaContent(sessionID, mediaID string) (*media.Media, error
 	}
 
 	media := media.NewMedia(contentTypes[0], responseBytes)
-	return &media, err
+
+	return media, err
 }
 
 // DeleteMediaContent deletes media related to a Yoti Doc Scan session based on the supplied media ID
--- a/docscan/session/retrieve/get_session_result.go
+++ b/docscan/session/retrieve/get_session_result.go
@@ -2,7 +2,6 @@ package retrieve
 
 import (
 	"encoding/json"
-	"fmt"
 
 	"github.com/getyoti/yoti-go-sdk/v3/docscan/constants"
 )
@@ -62,9 +61,6 @@ func (g *GetSessionResult) UnmarshalJSON(data []byte) error {
 
 		case constants.Liveness:
 			g.livenessChecks = append(g.livenessChecks, &LivenessCheckResponse{CheckResponse: check})
-
-		default:
-			fmt.Printf("Unrecognized check type: `%s`", check.Type)
 		}
 	}
 
--- a/docscan/session/retrieve/get_session_result.go
+++ b/docscan/session/retrieve/get_session_result.go
@@ -1,12 +1,72 @@
 package retrieve
 
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/getyoti/yoti-go-sdk/v3/docscan/constants"
+)
+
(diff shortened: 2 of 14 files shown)


docscan/client.go
```diff
@@ -153,7 +153,7 @@ func (c *Client) DeleteSession(sessionID string) error {
 }
 
 // GetMediaContent retrieves media related to a Yoti Doc Scan session based on the supplied media ID
-func (c *Client) GetMediaContent(sessionID, mediaID string) (*media.Media, error) { // TODO: change to media.Value
+func (c *Client) GetMediaContent(sessionID, mediaID string) (*media.Media, error) {
```
lkramer, line 156: since `media.Media` is an interface, you should not return it as a pointer.

docscan/session/retrieve/get_session_result.go
```diff
@@ -1,12 +1,72 @@
...
+			g.faceMatchChecks = append(g.faceMatchChecks, &FaceMatchCheckResponse{CheckResponse: check})
+
+		case constants.IDDocumentTextDataCheck:
+			g.textDataChecks = append(g.textDataChecks, &TextDataCheckResponse{CheckResponse: check})
+
+		case constants.Liveness:
+			g.livenessChecks = append(g.livenessChecks, &LivenessCheckResponse{CheckResponse: check})
+
+		default:
+			fmt.Printf("Unrecognized check type: `%s`", check.Type)
```
lkramer, line 67: It is my understanding that it's considered bad form for an SDK to write directly to std out. I'm not really sure what to do in this case, it might be safer to either return an error, or ignore the case. Since it's possible to get all the checks even if this fails, I'm leaning towards just ignoring.

davidgrayston: Other parts of the SDK use `log.Printf`, so perhaps `fmt.Printf` wasn't intended here?

If this is the case, does it make sense to use `log.Printf` here (and in other parts of SDK)? If not, perhaps we should review it use throughout the SDK as a separate task?

lkramer: Opinions may vary here, but I would not log in the SDK.

Integrators could be using std err and std out for their own logging and mandate certain formatting that would then be screwed up by the SDK coming along and do it's own thing.

I know I have struggled with similar things in other projects.

lkramer: One way around could be to support a custom logger interface which integrators can use if they want more detailed information, but it might be overkill in this case. You're probably better off doing a best effort in handling unexpected situations, and if you fail return an error.

echarrod: I'm unsure if we want to be returning an error in this instance though if there's an unrecognised check. If we add a new check in future, we don't want to stop clients from being able to process the other checks, which might happen if we throw an error.

On the other hand, they surely would have configured this check to come through, so maybe we should be throwing an error, since they should be expecting it.
Writing this, I think I'm leaning towards that we should be returning an error.

echarrod: Regarding the logging, I guess you could argue we should always be returning that information as an error when possible. So we can revisit that as a separate task

davidgrayston: In other SDKs, we just return a generic `CheckResponse`, e.g.
- https://github.com/getyoti/yoti-java-sdk/blob/master/yoti-sdk-impl/src/main/java/com/yoti/api/client/docs/session/retrieve/SimpleCheckResponse.java#L16
- https://github.com/getyoti/yoti-php-sdk/blob/master/src/DocScan/Session/Retrieve/GetSessionResult.php#L175

davidgrayston: I think we should skip without logging or returning an error. An new/unknown check will be available in `Checks` and will have a `Type`

lkramer: You could do a Generic Check Response, but you in a way already have that in that you expose the Checks slice which an SDK user can defer to if they have a check that is not supported in their version of the SDK.

echarrod: Yes they can get their check from the generic Checks slice, so we'll be fine. I've made an update in 1a2d3f9 to remove this logging. We'll look at the logging in other places in the SDK before we release the major

PR #168, opened 2020-08-18
SDK-1300: pointers

echarrod:
- Use pointers for doc scan structs & receivers
- Don't use pointer for time.Time - from [godocs](https://golang.org/src/time/time.go): 
> Programs using times should typically store and pass them as values, not pointers. That is, time variables and struct fields should be of type time.Time, not *time.Time.

### DIFF
--- a/docscan/session/retrieve/check_response.go
+++ b/docscan/session/retrieve/check_response.go
@@ -12,8 +12,8 @@ type CheckResponse struct {
 	ResourcesUsed  []string          `json:"resources_used"`
 	GeneratedMedia []*GeneratedMedia `json:"generated_media"`
 	Report         *ReportResponse   `json:"report"`
-	Created        time.Time         `json:"created"`
-	LastUpdated    time.Time         `json:"last_updated"`
+	Created        *time.Time        `json:"created"`
+	LastUpdated    *time.Time        `json:"last_updated"`
 }
 
 // AuthenticityCheckResponse represents a Document Authenticity check for a given session
--- a/docscan/session/retrieve/check_response.go
+++ b/docscan/session/retrieve/check_response.go
@@ -12,8 +12,8 @@ type CheckResponse struct {
 	ResourcesUsed  []string          `json:"resources_used"`
 	GeneratedMedia []*GeneratedMedia `json:"generated_media"`
 	Report         *ReportResponse   `json:"report"`
-	Created        *time.Time        `json:"created"`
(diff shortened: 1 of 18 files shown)


echarrod:
@lkramer do you know if we can always expect a check/task/media to have a createdDate and lastUpdated date? Just wondering whether we want to have pointers for values like this, where we can have nil if for some reason they are not present

davidgrayston:
I think it would be safer (less misleading) for dates to be `nil` instead of `0001-01-01T00:00:00Z`, when not present. Other SDKs are `nil/null` in this unlikely situation.

davidgrayston (approved):
Look good - I'm leaning towards date/times being pointers so that they are `nil` when not available (based on what we have done in other SDKs). If we are certain they will always be present, then it's fine how you have it.

lkramer:
> @lkramer do you know if we can always expect a check/task/media to have a createdDate and lastUpdated date? Just wondering whether we want to have pointers for values like this, where we can have nil if for some reason they are not present

The "correct" way to check if a time object is 0 is to use IsZero() https://golang.org/pkg/time/#Time.IsZero

docscan/session/retrieve/check_response.go
```diff
@@ -6,14 +6,14 @@ import (
...
-	Created        *time.Time       `json:"created"`
-	LastUpdated    *time.Time       `json:"last_updated"`
+	ID             string            `json:"id"`
+	Type           string            `json:"type"`
+	State          string            `json:"state"`
+	ResourcesUsed  []string          `json:"resources_used"`
+	GeneratedMedia []*GeneratedMedia `json:"generated_media"`
+	Report         *ReportResponse   `json:"report"`
+	Created        time.Time         `json:"created"`
+	LastUpdated    time.Time         `json:"last_updated"`
```
davidgrayston, line 16: We might need to add `omitempty` to properties if they're not always present. I believe we have done this in all other SDKs.

davidgrayston (changes requested):

docscan/session/retrieve/check_response.go
davidgrayston: As discussed `omitempty` isn't necessary when unserializing

echarrod:
Have gone with using pointers in e75ee06, based on what we have in the other SDKs, and also wanting to be as clear as possible that the value isn't present, rather than it potentially being missed if the time.Time was just the default value

davidgrayston (approved):

PR #174, opened 2020-08-25
SDK-1618: Include accept_self_asserted false in JSON payload

davidgrayston:
### Fixed
- `"accept_self_asserted": false` is now included in the payload

### DIFF
--- a/dynamic/wanted_attribute_builder.go
+++ b/dynamic/wanted_attribute_builder.go
@@ -17,7 +17,7 @@ type WantedAttributeBuilder struct {
 
 // WantedAttribute represents a wanted attribute in a dynamic sharing policy
 type WantedAttribute struct {
-	Name               string                `json:"name,omitempty"`
+	Name               string                `json:"name"`
 	Derivation         string                `json:"derivation,omitempty"`
 	Constraints        []constraintInterface `json:"constraints,omitempty"`
 	AcceptSelfAsserted *bool                 `json:"accept_self_asserted,omitempty"`
--- a/dynamic/wanted_attribute_builder.go
+++ b/dynamic/wanted_attribute_builder.go
@@ -1,7 +1,6 @@
 package dynamic
 
 import (
-	"encoding/json"
 	"errors"
 )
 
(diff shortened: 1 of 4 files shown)


dynamic/wanted_attribute_builder.go
```diff
@@ -18,61 +17,46 @@ type WantedAttributeBuilder struct {
 
 // WantedAttribute represents a wanted attribute in a dynamic sharing policy
 type WantedAttribute struct {
-	name               string
-	derivation         string
-	constraints        []constraintInterface
-	acceptSelfAsserted bool
+	Name               string                `json:"name,omitempty"`
```
echarrod, line 20: I know it's just copied from the MarshallJSON method which we've now removed, but since we check that `Name` isn't null when building, I don't think it also makes sense to also `omitempty` here. Do you think we should remove this?

davidgrayston: Fixed in 32b9cf5

dynamic/wanted_attribute_builder.go
```diff
@@ -18,61 +17,46 @@ type WantedAttributeBuilder struct {
...
 }
 
 // WithConstraint adds a constraint to a wanted attribute
 func (builder *WantedAttributeBuilder) WithConstraint(constraint constraintInterface) *WantedAttributeBuilder {
-	builder.attr.constraints = append(builder.attr.constraints, constraint)
+	builder.attr.Constraints = append(builder.attr.Constraints, constraint)
 	return builder
 }
 
 // WithAcceptSelfAsserted enables self-asserted user details such as from Aadhar
```
echarrod, line 44:
```suggestion
// WithAcceptSelfAsserted enables self-asserted user details, such as those from Aadhar
```

echarrod (approved):
Looks good. For the values which we don't want to include in the JSON payload if they're not set, I am happy to keep as pointers, rather than converting them all to a tristate value. This keeps it more concise, is largely hidden from the end-users, and fits with what we've done in the rest of the SDK.

PR #179, opened 2020-08-27
SDK-1728: Errors defined in central location

echarrod:
We have some errors which:
1) have dynamic information
2) we want to collect multiple errors (before returning to the user)
3) we want to declare them as temporary

But for the simple errors which don't fall under these 3 categories, I have moved them to a central location, so integrators can check against them, without worrying about whether we'll change the error message. We wouldn't do this in a minor anyway, but it's still a good idea to do 😄 

Also took David's idea of changing `yotierror.New` to `yotierror.NewResponseError`, so we can easily distinguish between the standard error.New function, and indicate that it also has a response field on the error


yotierror/activitydetails.go
```diff
@@ -0,0 +1,14 @@
+package yotierror
+
+import "errors"
+
+var (
+	// InvalidTokenError means that the token used to call GetActivityDetails is invalid. Make sure you are retrieving this token correctly.
+	InvalidTokenError = errors.New("invalid Token")
```
davidgrayston, line 7: Not sure if this was intentional? (`TokenDecryptError` is lowercase _"token"_)
```suggestion
	InvalidTokenError = errors.New("invalid token")
```

echarrod: Ah, no that wasn't intentional, fixed in a46d0ac

davidgrayston (approved):

klaidas (approved):

PR #181, opened 2020-09-01
SDK-1729: Fix profile example

echarrod:
This fix is just to not assume full name is present in the template, and wrap it in an IF statement.

This issue was being hidden by another issue - we were panicking instead of gracefully handling the error. This meant the request was retried with the same token, which had then expired as it had been used once. 

I have updated the panic instances to redirect to the error page with the error information, (where we're not already trying to parse the error page)


davidgrayston (approved):
Ran profile and dynamic share examples 👍

PR #183, opened 2020-09-09
SDK-1556: Add doc scan demo

echarrod:
Using Gin - https://github.com/gin-gonic/gin


docscan/session/create/check/liveness_test.go
```diff
@@ -25,6 +25,7 @@ func ExampleRequestedLivenessCheckBuilder() {
 
 func TestRequestedLivenessCheckBuilder_MaxRetriesIsOmittedIfNotSet(t *testing.T) {
 	check, err := NewRequestedLivenessCheckBuilder().
+		ForZoomLiveness().
```
davidgrayston, line 28: Should `ForZoomLiveness()` be a separate example?

docscan/client.go
```diff
@@ -238,6 +256,22 @@ func (c *Client) GetSupportedDocuments() (*supported.DocumentsResponse, error) {
 	return &result, err
 }
 
+func (c *Client) validateParameters(sessionID string, mediaID string) error {
+	if sessionID == "" {
+		return errors.New("sessionID cannot be an empty string")
+	}
```
davidgrayston, line 262: Would it make sense to have a validate method for each parameter? e.g. `validateSessionID`, `validateMediaID`, which can be reused individually. This would mean we don't need to duplicate `sessionID` validation.

Perhaps a generic `validateID(name string, value string)` function would be suitable, where you provide the property name (to include in the error message)?

echarrod: Yes I was unsure about this. If I extracted them out I would have:
```go
	err := validateString("sessionID", sessionID)
	if err != nil {
		return nil, err
	}
	
	err = validateString("mediaID", mediaID)
	if err != nil {
		return nil, err
	}

	err = validateString("c.SdkID", c.SdkID)
	if err != nil {
		return nil, err
	}
```
And then if we're doing that, I actually think I prefer this:
```go
	if sessionID == "" {
		return nil, errors.New("sessionID must not be an empty string")
	}

	if mediaID == "" {
		return nil, errors.New("mediaID must not be an empty string")
	}
	
	if c.SdkID == "" {
		return nil, errors.New("c.SdkID must not be an empty string")
	}
```
As it's clearer what's going on.

So maybe i should revert back to that?

docscan/session/create/check/liveness_test.go
echarrod: I just added this because the liveness check isn't valid if it doesn't have a type. So we're not testing ZoomLiveness here, we're just making it a more valid example

docscan/client.go
davidgrayston: Yeah, I think it's clearer to repeat this. Could `%s must not be an empty string` be a constant?

echarrod: Could do. Would be harder for clients to compare against. But I don't think this is something they'll be comparing against

davidgrayston: ok, I think it's fine to revert to what you had

docscan/session/create/check/liveness_test.go
echarrod: Sorry, I misunderstood. Fixed in bb85135

davidgrayston (approved):
Ran against staging without any problems 👍

PR #187, opened 2020-09-16
SDK-1300: Add Doc Scan test coverage

davidgrayston:
### Added
- Added Doc Scan test coverage

### Fixed
- `client.GetMediaContent()` content type validation - now returns error when content type is not available.


echarrod (approved):
👌

PR #188, opened 2020-09-21
IN-3120: go Example QR

nikhilPank:


echarrod (approved):

davidgrayston (changes requested):
Do we need to make this change in the dynamic share example too?

https://github.com/getyoti/yoti-go-sdk/blob/development/_examples/profile/dynamic-share.html#L77

davidgrayston (approved):

PR #190, opened 2020-10-01
SDK-1733: Add manual fallback option to Doc Auth check

echarrod:
### Added
- Ability to configure manual fallback value on Document Authenticity Check

### DIFF
--- a/docscan/session/create/check/document_authenticity_test.go
+++ b/docscan/session/create/check/document_authenticity_test.go
@@ -21,6 +21,19 @@ func ExampleRequestedDocumentAuthenticityCheckBuilder() {
 	// Output: {"type":"ID_DOCUMENT_AUTHENTICITY","config":{"manual_check":"FALLBACK"}}
 }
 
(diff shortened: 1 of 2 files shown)


docscan/session/create/check/document_authenticity_test.go
```diff
@@ -3,17 +3,77 @@ package check
...
 	"encoding/json"
 	"fmt"
+	"testing"
+
+	"gotest.tools/v3/assert"
 )
 
 func ExampleRequestedDocumentAuthenticityCheckBuilder() {
-	check, err := NewRequestedDocumentAuthenticityCheckBuilder().Build()
+	docAuthCheck, err := NewRequestedDocumentAuthenticityCheckBuilder().WithManualCheckFallback().Build()
```
davidgrayston, line 12: Should we also have coverage without manual fallback config?

davidgrayston (approved):

docscan/session/create/check/document_authenticity_test.go
echarrod: Added in 1e55c77

PR #191, opened 2020-10-06
SDK-1740: Biometric Consent

davidgrayston:
#### Doc Scan
- Ability to block the collection of biometric consent


echarrod (approved):
Do you think we need to add a test that `block_biometric_consent` is omitted when not specified? 

Also I think "biometric" is just one word (in the change notes)

davidgrayston:
That's covered by `ExampleSessionSpecificationBuilder_Build` but we could add an extra test specifically for that?

Updated change notes (it was copied from other releases, so we might need to fix those too)

PR #192, opened 2020-10-08
SDK-1640: Chip Data

davidgrayston:
### Added
- Sandbox support for setting document ID photo

Note: SDK already allows chip data to be ignored/desired and parses ID photo in response

### DIFF
--- a/docscan/sandbox/request/task/document_text_data_extraction_task_test.go
+++ b/docscan/sandbox/request/task/document_text_data_extraction_task_test.go
@@ -14,7 +14,7 @@ func ExampleDocumentTextDataExtractionTaskBuilder() {
 		return
 	}
 
-	check, err := NewDocumentTextDataExtractionTaskBuilder().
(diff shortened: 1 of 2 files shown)


docscan/sandbox/request/task/document_text_data_extraction_task_test.go
```diff
@@ -14,7 +14,7 @@ func ExampleDocumentTextDataExtractionTaskBuilder() {
 		return
 	}
 
-	check, err := NewDocumentTextDataExtractionTaskBuilder().
+	task, err := NewDocumentTextDataExtractionTaskBuilder().
```
echarrod, line 17: Some linters complain if variable names are the same as package names in the project. But since we are within the `task` package, I think this is fine? So e.g. having `filter` here is not recommended, since we import `"github.com/getyoti/yoti-go-sdk/v3/docscan/sandbox/request/filter"`

echarrod (approved):

PR #208, opened 2020-12-14
SDK-1854: Add github actions

emmas-yoti:
Adds github actions to this repository

### DIFF
--- a/.github/workflows/sonar.yaml
+++ b/.github/workflows/sonar.yaml
@@ -1,0 +1,23 @@
+name: Sonar Scan
+on: [push, pull_request_target]
+
+jobs:
+  sonar:
+    name: Sonar Scan
+    runs-on: ubuntu-latest
+    # always run on push events
+    # only run on pull_request_target event when pull request pulls from fork repository
+    if: >
+      github.event_name == 'push' || 
+      github.event_name == 'pull_request_target' && github.event.pull_request.head.repo.full_name != github.repository 
+
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+
+      - uses: sonarsource/sonarcloud-github-action@master
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
+
--- a/.github/workflows/sonar.yaml
+++ b/.github/workflows/sonar.yaml
@@ -16,6 +16,14 @@ jobs:
         with:
           fetch-depth: 0
 
+      - uses: actions/setup-go@v2
+        with:
+          go-version: 1.15
+
+      - run: go test -json ./... > report.json
+
+      - run: go test -coverprofile=coverage.out -json ./... > sonar-report.json
+
       - uses: sonarsource/sonarcloud-github-action@master
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -1,0 +1,29 @@
+name: Unit Tests
+on: [push, pull_request_target]
+
+jobs:
+  tests:
+    name: Tests (Go ${{ matrix.go-version }})
+    runs-on: ubuntu-latest
+    # always run on push events
+    # only run on pull_request_target event when pull request pulls from fork repository
+    if: >
+      github.event_name == 'push' || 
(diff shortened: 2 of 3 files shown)


.github/workflows/sonar.yaml
```diff
@@ -0,0 +1,23 @@
...
+    # always run on push events
+    # only run on pull_request_target event when pull request pulls from fork repository
+    if: >
+      github.event_name == 'push' || 
+      github.event_name == 'pull_request_target' && github.event.pull_request.head.repo.full_name != github.repository 
+
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
```
echarrod, line 17: Presume we need this to fetch all of the commits? Looks like default is 1

.github/workflows/tests.yaml
```diff
@@ -0,0 +1,29 @@
...
+        go-version: [1.11, 1.12, 1.13, 1.14, 1.15, "^1"]
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: actions/setup-go@v2
+        with:
+          go-version: ${{ matrix.go-version }}
+
+      - run: ./sh/gofmt.sh
```
echarrod, line 25: Are we sure that this will fail the action if it's not formatted correctly? Doesn't seem to give us much info. in the output on this action when it's run. Same with `go vet`

.github/workflows/sonar.yaml
```diff
@@ -0,0 +1,23 @@
...
+
+      - uses: sonarsource/sonarcloud-github-action@master
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
+
```
echarrod, line 23: There's [no coverage information on this PR](https://sonarcloud.io/dashboard?branch=sdk-1854-github-actions&id=getyoti%3Ago), as there's no Go code. Would be good to check that the coverage works, but I guess the best way of doing that would to see how it runs on the next PR that modifies Go code

.github/workflows/sonar.yaml
emmas-yoti: Specifically, it's to make sure all the blame information is available to Sonar. Further information is here: https://community.sonarsource.com/t/sonarcloud-with-github-actions/22713

.github/workflows/tests.yaml
emmas-yoti: AFAIK both tools return a shell exit code of 1 if they fail, which the run step will check for.

echarrod (changes requested):
Good to go once we add the Github Actions badge - we can revisit the coverage information in a future PR when we have Go code in the PR to use it on

README.md
```diff
@@ -1,7 +1,7 @@
 # Yoti Go SDK
 
+[![Build Status](https://github.com/getyoti/yoti-go-sdk/workflows/Unit%20Test/badge.svg?branch=master)](https://github.com/getyoti/yoti-go-sdk/actions)
```
echarrod, line 3: The badge needs updating still:
```suggestion
[![Build Status](https://github.com/getyoti/yoti-go-sdk/workflows/Unit%20Tests/badge.svg?branch=master)](https://github.com/getyoti/yoti-go-sdk/actions)
```

emmas-yoti: I thought I committed this suggestion already?

.github/workflows/tests.yaml
```diff
@@ -0,0 +1,29 @@
...
+    runs-on: ubuntu-latest
+    # always run on push events
+    # only run on pull_request_target event when pull request pulls from fork repository
+    if: >
+      github.event_name == 'push' || 
+      github.event_name == 'pull_request_target' && github.event.pull_request.head.repo.full_name != github.repository 
+    strategy:
+      fail-fast: false
+      matrix:
+        go-version: [1.11, 1.12, 1.13, 1.14, 1.15, "^1"]
```
echarrod, line 16: [Go only supports the last two major releases](https://golang.org/doc/devel/release.html#:~:text=checkout%20goX.Y.Z-,Release%20Policy,until%20the%20Go%201.8%20release.&text=1%2C%20Go%201.6.,2%2C%20and%20so%20on).), so I think we're fine only running it against a few latest ones here (and we can add 1.16 too):
```suggestion
        go-version: [1.14, 1.15, 1.16, "^1"]
```

echarrod (changes requested):
I think we can drop some older versions, looks good otherwise

echarrod (approved):
Looks good now

PR #226, opened 2022-05-15
SDK-2088,SDK-2078: Add Identity Profile, Subject to Dynamic Scenario generation

echarrod:
### Changed
- Run instructions for Profile example

### Added
- Add subject and identityProfileRequirements to dynamic scenario generation, callable with:
```go
identityProfile := []byte(`{
	"trust_framework": "UK_TFIDA",
	"scheme": {
		"type":      "DBS",
		"objective": "BASIC"
	}
}`)

policy, err := (&dynamic.PolicyBuilder{}).
	WithIdentityProfileRequirements(identityProfile).
	Build()
if err != nil {
	return err
}

subject := []byte(`{
	"subject_id": "my_subject_id"
}`)
scenario, err := (&dynamic.ScenarioBuilder{}).
	WithPolicy(policy).
	WithSubject(subject).
	WithCallbackEndpoint(yourEndpoint).Build()
```
Example available when running the profile example project at <https://localhost:8080/dbs-check>:
image


_examples/profile/.gitignore
```diff
@@ -3,3 +3,6 @@
 # Example project generated self-signed certificate
 /yotiSelfSignedCert.pem
 /yotiSelfSignedKey.pem
+
+# Compiled binary
+/profile
```
Alttaf, line 8: New line

echarrod: Added in https://github.com/getyoti/yoti-go-sdk/pull/226/commits/a072aadfae3cd4d62a949e0ae7bfee63c9b06960

_examples/profile/dynamicshare.go
```diff
@@ -0,0 +1,96 @@
...
+	if err != nil {
+		errorPage(w, req.WithContext(context.WithValue(
+			req.Context(),
+			contextKey("yotiError"),
+			fmt.Sprintf("%s", err.Error()),
+		)))
+		return
+	}
+
+	templateVars := map[string]interface{}{
```
Alttaf, line 70: why the map string interface, could you use a known struct for this?

Alttaf (approved):

jveling-yoti (approved):

_examples/profile/dynamicshare.go
echarrod: I'm just moving this file, but it does appear in all our templates. 

I don't think it's worth creating a struct for these, since we would have to create 4/5 different structs which would all be used once,  and we don't gain the benefits of using a struct when we reference this object in the template: we just do e.g. `{{.pageTitle}}` (with no autocomplete), and the template rendering engine picks it up at runtime

PR #227, opened 2022-05-15
SDK-1536,SDK-2073 Update protobuf files

echarrod:
### Changed
- Protobuf files after regeneration (for Identity Profile requirements)
- use "google.golang.org/protobuf" instead of "github.com/golang/protobuf/proto" (latter is deprecated)


jveling-yoti (approved):

jveling-yoti:
Where are the actual .proto files kept?

echarrod:
> Where are the actual .proto files kept?

There are kept private, in the sdk-protobuf repo. We don't want to expose these to the public

PR #229, opened 2022-05-18
SDK-2083: Identity Profile Report attribute

echarrod:
### Added
- Identity Profile Report attribute:
```go
var identityProfileReportValue map[string]interface{} = userProfile.IdentityProfileReport().Value() // JSON
```

### DIFF
--- a/_examples/profile/profile.go
+++ b/_examples/profile/profile.go
@@ -3,6 +3,7 @@ package main
 import (
 	"bytes"
 	"context"
+	"encoding/json"
 	"fmt"
 	"html/template"
 	"image"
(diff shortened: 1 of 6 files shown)


BenSmithYoti (approved):
LGTM.

_examples/profile/profile.go
```diff
@@ -103,6 +104,13 @@ func profile(w http.ResponseWriter, r *http.Request) {
 					prevalue,
 				}
 			},
+			"jsonMarshallIndent": func(data interface{}) string {
+				json, err := json.MarshalIndent(data, "", "\t")
+				if err != nil {
+					fmt.Println(err)
```
Alttaf, line 110: if this errors you will still return the string(json). is this intentional?

Alttaf (approved):
just a comment worth tripple checking

_examples/profile/profile.go
echarrod: Good question. This function is used inside the template, and since it's rendered HTML, there's not an easy way to handle the errors

PR #233, opened 2022-05-19
[SDK-2144] - passing sharing error response from connect to integrator

jveling-yoti:
### Added

A more detailed error response when sharing is not successful.

```
type DetailedSharingFailureError struct {
	Code        string
	Description string
}
```

This enables the integrator to check the error type `err.Is(DetailedSharingFailureError)`


profile/service.go
```diff
@@ -132,3 +132,14 @@ func parseIsAgeVerifiedValue(byteValue []byte) (result *bool, err error) {
 
 	return
 }
+
+func handleUnsuccessfulShare(parsedResponse profileDO) error {
+	if parsedResponse.ErrorDetails != nil {
+		return yotierror.DetailedSharingFailureError{
+			Code:        parsedResponse.ErrorDetails.ErrorCode,
```
BenSmithYoti, line 139: Is it possible that there may be non nil ErrorDetails but nil 
`parsedResponse.ErrorDetails.ErrorCode`
or 
`parsedResponse.ErrorDetails.Description`
?

BenSmithYoti (approved):
One question other than that LGTM.

profile/service.go
jveling-yoti: Good question - I dont believe so, but def worth double checking. Ill take a look

BenSmithYoti: Yeah my understanding is that for the specific use case this is being added for these KVPs _will_ be present, but there _may possibly_ be other cases where there's no error_code at least. But I am not sure. I believe Paolo and/or Joe would know.

BenSmithYoti: I see the commit to update this - thanks - am wondering though, dotnet side we capture the whole of the Response content whenever `SharingOutcome != "SUCCESS"` to cover cases where the above conditions are not met. 

This provides visibility of the entire response Json _regardless of the presence/absence of particular key value pairs_. 

`ResponseContent = responseContent;`

https://github.com/getyoti/yoti-dotnet-sdk/blob/f9fda98dce74bf89e33744b718449f83db7bd1f5/src/Yoti.Auth/Exceptions/YotiProfileException.cs

Perhaps you want to do the same here?

PR #253, opened 2023-01-04
SDK-2188-go-support-for-expired-documents-at-session-creation

mehmet-yoti:
## IDV

**Added**

- Support for expired documents at session creation

**Requested with**
```go	
passportFilter, err := filter.NewRequestedOrthogonalRestrictionsFilterBuilder().
		WithIncludedDocumentTypes(
			[]string{"PASSPORT"}).
		WithExpiredDocuments(true).
		Build()
	if err != nil {
		return nil, err
	}
```


fofiuiancu (changes requested):
- [_examples/idv/idv] was added when it should have been (git) ignored
- I see WithExpiredDocuments being used but not where it has been added

mehmet-yoti:
missing files were added. can we recheck

fofiuiancu (approved):

PR #254, opened 2023-01-04
Sdk 2037 go support for non latin documents at session creation

mehmet-yoti:
## IDV

**Added**

- Support for non latin documents at session creation

**Requested with**
```go	passportFilter, err := filter.NewRequestedOrthogonalRestrictionsFilterBuilder().
		WithIncludedDocumentTypes(
			[]string{"PASSPORT"}).
		WithExpiredDocuments(true).
		WithNonLatinDocuments(true).
		Build()
	if err != nil {
		return nil, err
	}
```


fofiuiancu (changes requested):
it seems that this PR also includes tha changes for allowExpiredDocuments?

docscan/session/create/filter/orthogonal_restrictions_filter.go
```diff
@@ -76,10 +86,22 @@ func (b *RequestedOrthogonalRestrictionsFilterBuilder) WithExcludedDocumentTypes
 	return b
 }
 
+func (b *RequestedOrthogonalRestrictionsFilterBuilder) WithNonLatinDocuments(allowNonLatinDocuments bool) *RequestedOrthogonalRestrictionsFilterBuilder {
```
fofiuiancu, line 89: missing comment

docscan/session/create/filter/orthogonal_restrictions_filter.go
```diff
@@ -86,11 +86,13 @@ func (b *RequestedOrthogonalRestrictionsFilterBuilder) WithExcludedDocumentTypes
 	return b
 }
 
+// WithNonLatinDocuments sets a bool value to allowNonDocuments on filter
```
fofiuiancu, line 89:
```suggestion
// WithNonLatinDocuments sets a bool value to allowNonLatinDocuments on filter
```

fofiuiancu (approved):

PR #261, opened 2023-02-01
Added support for ImportToken

fofiuiancu:
**Added**

- ImportToken to Session Specification Builder and a ImportTokenBuilder

**Requested with**
```go	
ttl := time.Hour * 24 * 30
importToken, err := create.NewImportTokenBuilder().
	WithTTL(int(ttl.Seconds())).
	Build()
if err != nil {
	return nil, err
}

sessionSpec, err = create.NewSessionSpecificationBuilder().
		WithClientSessionTokenTTL(6000).
		WithResourcesTTL(900000).
		WithUserTrackingID("some-tracking-id").
		WithSDKConfig(sdkConfig).
		WithIdentityProfileRequirements(identityProfile).
		WithCreateIdentityProfilePreview(true).
		WithSubject(subject).
                WithImportToken(importToken).
		Build()
```

- ImportTokenResponse to IDV GetSessionResult:

```go
getSessionResult, err := client.GetSession(sessionId)
if err != nil {
        return err
}
importTokenResponse := getSessionResult.ImportTokenResponse
```

Example available when running the idv example project at https://localhost:8080/dbs

### DIFF
--- a/docscan/session/create/import_token.go
+++ b/docscan/session/create/import_token.go
@@ -7,7 +7,7 @@ type ImportToken struct {
 }
 
 // defaultImportTokenTTL is 12 Months
-const defaultImportTokenTTL = time.Hour * 24 * 30 * 12
+const defaultImportTokenTTL = time.Hour * 24 * 365
 
 // NewImportTokenBuilder creates a new ImportTokenBuilder
 func NewImportTokenBuilder() *ImportTokenBuilder {
--- a/docscan/session/create/import_token.go
+++ b/docscan/session/create/import_token.go
@@ -1,0 +1,35 @@
+package create
+
+import "time"
+
+type ImportToken struct {
+	Ttl int `json:"ttl"`
+}
+
+// defaultImportTokenTTL is 12 Months
+const defaultImportTokenTTL = time.Hour * 24 * 30 * 12
+
+// NewImportTokenBuilder creates a new ImportTokenBuilder
+func NewImportTokenBuilder() *ImportTokenBuilder {
+	return &ImportTokenBuilder{
+		Ttl: int(defaultImportTokenTTL.Seconds()),
+	}
+}
+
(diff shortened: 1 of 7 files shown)


docscan/session/create/import_token.go
```diff
@@ -0,0 +1,35 @@
+package create
+
+import "time"
+
+type ImportToken struct {
+	Ttl int `json:"ttl"`
+}
+
+// defaultImportTokenTTL is 12 Months
+const defaultImportTokenTTL = time.Hour * 24 * 30 * 12
```
klaidas, line 10: 12 months seems like a very long time, i'd make the default 1 month so we're recommending security (by giving an attacker less time) and only the RP can change it if they need it to be longer.

mehmet-yoti (approved):

docscan/session/create/session_spec.go
```diff
@@ -140,6 +144,12 @@ func (b *SessionSpecificationBuilder) WithSubject(subject json.RawMessage) *Sess
 	return b
 }
 
+// WithImportToken sets whether of not an ImportToken is to be generated.
```
klaidas, line 147:
```suggestion
// WithImportToken sets whether or not an ImportToken is to be generated.
```

klaidas (approved):
LGTM

docscan/session/create/import_token.go
klaidas: oh also i wouldn't say that's accurately 12 months, if you wanna do a year i would've just done `time.Hour * 24 * 365` (rip leap years)

fofiuiancu: in the ticket we're defining ttl as: "seconds, not null, default to 12 months, MAX 12 months, not less than 30 days"

klaidas (approved):

PR #271, opened 2023-04-28
SDK-2304:Added expired document feature to document restriction

mehmet-yoti:
Added Support for expired documents at session creation using DOCUMENT_RESTRICTIONS

### DIFF
--- a/docscan/session/create/filter/document_restrictions_filter.go
+++ b/docscan/session/create/filter/document_restrictions_filter.go
@@ -20,7 +20,7 @@ func (r *RequestedDocumentRestrictionsFilter) MarshalJSON() ([]byte, error) {
 		Type                  string                          `json:"type"`
 		Inclusion             string                          `json:"inclusion"`
 		Documents             []*RequestedDocumentRestriction `json:"documents"`
-		AllowExpiredDocuments *bool                           `json:"allow_expired_documents"`
+		AllowExpiredDocuments *bool                           `json:"allow_expired_documents,omitempty"`
 	}{
 		Type:                  r.Type(),
 		Inclusion:             r.inclusion,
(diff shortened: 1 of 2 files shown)


docscan/session/create/filter/document_restrictions_filter.go
```diff
@@ -4,8 +4,9 @@ import "encoding/json"
 
 // RequestedDocumentRestrictionsFilter filters for a required document, allowing specification of restrictive parameters
 type RequestedDocumentRestrictionsFilter struct {
-	inclusion string
-	documents []*RequestedDocumentRestriction
+	inclusion             string
+	documents             []*RequestedDocumentRestriction
+	allowExpiredDocuments *bool
```
lkramer, line 9: Do we always use pointers here? Why?

docscan/session/create/filter/document_restrictions_filter.go
```diff
@@ -16,20 +17,23 @@ func (r RequestedDocumentRestrictionsFilter) Type() string {
 // MarshalJSON returns the JSON encoding
 func (r *RequestedDocumentRestrictionsFilter) MarshalJSON() ([]byte, error) {
 	return json.Marshal(&struct {
-		Type      string                          `json:"type"`
-		Inclusion string                          `json:"inclusion"`
-		Documents []*RequestedDocumentRestriction `json:"documents"`
+		Type                  string                          `json:"type"`
+		Inclusion             string                          `json:"inclusion"`
+		Documents             []*RequestedDocumentRestriction `json:"documents"`
+		AllowExpiredDocuments *bool                           `json:"allow_expired_documents,omitempty"`
```
lkramer, line 23: what is the purpose of using both `*bool` and `omitempty`?

docscan/session/create/filter/document_restrictions_filter.go
lkramer: I guess I understand, if the defaults are set on the backend then it can make sense.

mehmet-yoti: Yes you are right

docscan/session/create/filter/document_restrictions_filter.go
mehmet-yoti: Thats resolved with commit

lkramer (approved):
LGTM

PR #272, opened 2023-05-03
SDK-2303:added support-for-non-latin-documents-when-creating-a-sessio…

mehmet-yoti:
Added Support for non latin documents at session creation using DOCUMENT_RESTRICTIONS


lkramer (approved):
LGTM

PR #277, opened 2023-06-16
SDK-2334:added support for enabling expanded document fields

mehmet-yoti:


saurabh-yoti:
PR looks good. Should remove `idv.exe` executable from the `'_examples/idv'` folder.

PR #282, opened 2023-08-08
Sdk 2246 go retrieve session

mehmet-yoti:


digitalidentity/share_session.go
```diff
@@ -0,0 +1,31 @@
...
+)
+
+// ShareSessionResult contains a dynamic share QR code
+type ShareSessionResult struct {
+	Id      string  `json:"id"`
+	Status  string  `json:"status"`
+	Expiry  string  `json:"expiry"`
+	Created string  `json:"created"`
+	Updated string  `json:"updated"`
+	QrCode  qrCode  `json:"qrCode"`
```
fofiuiancu, line 19: I believe this and Receipt can be missing so lets have them as pointers

digitalidentity/service_test.go
```diff
@@ -4,13 +4,28 @@ import (
...
 	"net/http"
+	"net/http/httptest"
 	"strings"
 	"testing"
 
 	"github.com/getyoti/yoti-go-sdk/v3/test"
 	"gotest.tools/v3/assert"
 )
 
+/*
```
fofiuiancu, line 15: leftover

digitalidentity/service.go
```diff
@@ -20,26 +21,61 @@ type SessionResult struct {
...
 
+	headers := requests.AuthHeader(clientSdkId)
+
 	request, err := requests.SignedRequest{
 		Key:        key,
 		HTTPMethod: http.MethodPost,
 		BaseURL:    apiUrl,
 		Endpoint:   endpoint,
-		Headers:    nil,
+		Headers:    headers,
```
fofiuiancu, line 39:
```suggestion
		Headers:    requests.AuthHeader(clientSdkId),
```

digitalidentity/service.go
```diff
@@ -20,26 +21,61 @@ type SessionResult struct {
 }
 
 // CreateShareSession creates session using the supplied session specification
-func CreateShareSession(httpClient requests.HttpClient, shareSession *ShareSession, clientSdkId, apiUrl string, key *rsa.PrivateKey) (share ShareURL, err error) {
+func CreateShareSession(httpClient requests.HttpClient, shareSession *ShareSession, clientSdkId, apiUrl string, key *rsa.PrivateKey) (share SessionResult, err error) {
```
fofiuiancu, line 24:
```suggestion
func CreateShareSession(httpClient requests.HttpClient, shareSessionRequest *ShareSessionRequest, clientSdkId, apiUrl string, key *rsa.PrivateKey) (shareSession shareSession, err error) {
```

digitalidentity/service.go
```diff
@@ -20,26 +21,61 @@ type SessionResult struct {
...
+// GetSession get session info using the supplied sessionID
+func GetSession(httpClient requests.HttpClient, sessionID string, clientSdkId, apiUrl string, key *rsa.PrivateKey) (share ShareSessionResult, err error) {
+	endpoint := identitySesssionCreationEndpoint
+	headers := requests.AuthHeader(clientSdkId)
+	request, err := requests.SignedRequest{
+		Key:        key,
+		HTTPMethod: http.MethodGet,
+		BaseURL:    apiUrl,
+		Endpoint:   endpoint,
+		Headers:    headers,
```
fofiuiancu, line 73:
```suggestion
		Headers:    requests.AuthHeader(clientSdkId),
```

digitalidentity/service.go
```diff
@@ -20,26 +21,61 @@ type SessionResult struct {
...
+	}.Request()
+	if err != nil {
+		return share, err
+	}
+
```
fofiuiancu, line 78: can't comment on the line below, but ShareURLHTTPErrorMessages is not the correct mapping

digitalidentity/service.go
```diff
@@ -20,26 +21,61 @@ type SessionResult struct {
...
+	err = json.Unmarshal(responseBytes, &share)
+
+	return share, err
+}
+
+// GetSession get session info using the supplied sessionID
+func GetSession(httpClient requests.HttpClient, sessionID string, clientSdkId, apiUrl string, key *rsa.PrivateKey) (share ShareSessionResult, err error) {
+	endpoint := identitySesssionCreationEndpoint
```
fofiuiancu, line 66: sessionID param should be added here

digitalidentity/service_test.go
```diff
@@ -29,7 +44,7 @@ func ExampleCreateShareSession() {
 		do: func(*http.Request) (*http.Response, error) {
 			return &http.Response{
 				StatusCode: 201,
-				Body:       io.NopCloser(strings.NewReader(`{"qrcode":"https://code.yoti.com/CAEaJDQzNzllZDc0LTU0YjItNDkxMy04OTE4LTExYzM2ZDU2OTU3ZDAC","ref_id":"0"}`)),
+				Body:       io.NopCloser(strings.NewReader(`{"status":"success","ref_id":"0"}`)),
```
fofiuiancu, line 47: the response as described in https://lampkicking.atlassian.net/wiki/spaces/CON/pages/3756458012/Share+v2+Flows#API-Reference
should be:
{
  "id": "",
  "status": "",
  "expiry": ""
}

digitalidentity/service_test.go
```diff
@@ -90,7 +105,7 @@ func TestCreateShareURL_Unsuccessful_400(t *testing.T) {
 	assert.Check(t, !temporary || !tempError.Temporary())
 }
 
-func createShareUrlWithErrorResponse(statusCode int, responseBody string) (share ShareURL, err error) {
+func createShareUrlWithErrorResponse(statusCode int, responseBody string) (share SessionResult, err error) {
```
fofiuiancu, line 108: can't comment above
Let's add some Create/Fetch specific errors so we can test that mapping that was missing.

digitalidentity/service_test.go
```diff
@@ -113,3 +128,42 @@ func createShareUrlWithErrorResponse(statusCode int, responseBody string) (share
 
 	return CreateShareSession(client, &scenario, "sdkId", "https://apiurl", key)
 }
+
+type MockHttpClient struct{}
```
fofiuiancu, line 132: we don't really need this.
We can just use the same mechanism as above where teh mockClient has a do function that returns the desired payload
eg
do: func(*http.Request) (*http.Response, error) {
			return &http.Response{
				StatusCode: 201,
				Body:       io.NopCloser(strings.NewReader(`{"status":"success","ref_id":"0"}`)),
			}, nil
		},

fofiuiancu (changes requested):
Not covered by this PR but should have been covered by the previous PR:
The ShareSessionRequest should have the optional fields as pointers: 
(eg policy is mandatory in Share V2 so it shouldn't be a pointer)

Please see the optional ones here: 
https://lampkicking.atlassian.net/wiki/spaces/DAS/pages/3900801110/Share+V2+Payload+documentation

_examples/idv/routes.go
```diff
@@ -12,4 +12,5 @@ func initializeRoutes() {
 	router.GET("/success", showSuccessPage)
 	router.GET("/media", getMedia)
 	router.GET("/privacy-policy", showPrivacyPolicyPage)
+	router.GET("/digital-identity", showDigitalPage)
```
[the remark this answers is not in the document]
fofiuiancu: this should have been removed as well _examples/idv/models.sessionspec.go

fofiuiancu:
> Not covered by this PR but should have been covered by the previous PR: The ShareSessionRequest should have the optional fields as pointers: (eg policy is mandatory in Share V2 so it shouldn't be a pointer)
> 
> Please see the optional ones here: https://lampkicking.atlassian.net/wiki/spaces/DAS/pages/3900801110/Share+V2+Payload+documentation

this should also be addressed

fofiuiancu (changes requested):

digital_identity_client_test.go
```diff
@@ -44,18 +46,49 @@ func TestDigitalIDClient_CreateShareURL(t *testing.T) {
...
+			do: func(*http.Request) (*http.Response, error) {
+				return &http.Response{
+					StatusCode: 404,
+				}, nil
+			},
+		},
+		Key: key,
+	}
+
+	_, err := client.GetSession("SOME ID")
```
klaidas, line 87: let's add a success test case for this too

digitalidentity/service.go
```diff
@@ -33,14 +36,47 @@ func CreateShareSession(httpClient requests.HttpClient, shareSession *ShareSessi
...
+	}
+
+	err = json.Unmarshal(responseBytes, &share)
+
+	return share, err
+}
+
+// GetSession get session info using the supplied sessionID
+func GetSession(httpClient requests.HttpClient, sessionID string, clientSdkId, apiUrl string, key *rsa.PrivateKey) (share ShareSession, err error) {
+	endpoint := identitySesssionCreationEndpoint
```
klaidas, line 65: wrong endpoint, should be `identitySessionRetrieval` (and should substitute the `%s`)

```suggestion
	endpoint := fmt.Sprintf(identitySessionRetrieval, sessionID)
```

digitalidentity/service.go
```diff
@@ -10,20 +10,23 @@ import (
 	"github.com/getyoti/yoti-go-sdk/v3/yotierror"
 )
 
-const identitySesssionCreationEndpoint = "/v2/sessions"
+const identitySesssionCreationEndpoint = "v2/sessions"
```
klaidas, line 13:
```suggestion
const identitySessionCreationEndpoint = "v2/sessions"
```

typo

digitalidentity/service.go
```diff
@@ -10,20 +10,23 @@ import (
...
 
 // SessionResult contains the information about a created session
-type SessionResult struct {
+
+/*type SessionResult struct {
 	Id     int    `json:"id"`
 	Status string `json:"status"`
 	Expiry string `json:"expiry"`
 }
+*/
```
klaidas, line 23: don't need any of this by the looks of it, so can just delete

digitalidentity/service.go
```diff
@@ -33,14 +36,47 @@ func CreateShareSession(httpClient requests.HttpClient, shareSession *ShareSessi
...
+		HTTPMethod: http.MethodGet,
+		BaseURL:    apiUrl,
+		Endpoint:   endpoint,
+		Headers:    headers,
+	}.Request()
+	if err != nil {
+		return share, err
+	}
+
+	response, err := requests.Execute(httpClient, request, ShareSessionHTTPErrorMessages, yotierror.DefaultHTTPErrorMessages)
```
klaidas, line 78: i think we need to update this `requests` package to support the error codes connect send back, some of these are really useful and we should differentiate between them and give the RP an exported error var they can switch on and handle. 

e.g. from looking at [this](https://lampkicking.atlassian.net/wiki/spaces/CON/pages/3756458012/Share+v2+Flows) `404` `UNKNOWN_SESSION` is invalid session id which would be useful for a RP to know when integrating with sdk, as opposed to the `ShareSessionHTTPErrorMessages` where `404` is mapped to `Application was not found`?

digitalidentity/share_session.go
```diff
@@ -0,0 +1,31 @@
+package digitalidentity
+
+var (
+	// ShareURLHTTPErrorMessages specifies the HTTP error status codes used
+	// by the Share URL API
```
klaidas, line 5:
```suggestion
	// ShareSessionHTTPErrorMessages specifies the HTTP error status codes used
	// by the Share Session API
```

digitalidentity/share_session.go
```diff
@@ -0,0 +1,31 @@
...
+	ShareSessionHTTPErrorMessages = map[int]string{
+		400: "JSON is incorrect, contains invalid data",
+		404: "Application was not found",
+	}
+)
+
+// ShareSession object
```
klaidas, line 12:
```suggestion
// ShareSession contains information about the session. 
```

digitalidentity/share_session_builder.go
```diff
@@ -6,12 +6,12 @@ import (
 
 // ShareSessionBuilder builds a session
 type ShareSessionBuilder struct {
-	shareSession ShareSession
-	err          error
+	shareSessionRequest ShareSessionRequest
+	err                 error
 }
 
 // ShareSession represents a sharesession
```
klaidas, line 13: update the comment

digitalidentity/service.go
klaidas: @mehmet-yoti are you still working on this?

mehmet-yoti: Yes I am doing some changes on that.

fofiuiancu:
> > Not covered by this PR but should have been covered by the previous PR: The ShareSessionRequest should have the optional fields as pointers: (eg policy is mandatory in Share V2 so it shouldn't be a pointer)
> > Please see the optional ones here: https://lampkicking.atlassian.net/wiki/spaces/DAS/pages/3900801110/Share+V2+Payload+documentation
> 
> this should also be addressed

ShareSessionNotification should also be optional
and the Builder should take care of initialising the optionals with defaults where is the case (I a not sure why the extensions is populated in the Builder with an empty array, but haven't investigated it in depth)

digital_identity_client_test.go
```diff
@@ -34,28 +32,12 @@ func TestDigitalIDClient_KeyLoad_Failure(t *testing.T) {
...
-		do: func(*http.Request) (*http.Response, error) {
-			return &http.Response{
-				StatusCode: 201,
-				Body:       io.NopCloser(strings.NewReader(`{"qrcode":"https://code.yoti.com/some-qr","ref_id":"0"}`)),
-			}, nil
-		},
-	}
 
-	policy, err := (&dynamic.PolicyBuilder{}).WithFullName().WithWantedRememberMe().Build()
+	policy, err := (&digitalidentity.PolicyBuilder{}).WithFullName().WithWantedRememberMe().Build()
```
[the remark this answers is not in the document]
fofiuiancu: we need to add tests for the errors in  the newly expected format. TestDigitalIDClient_HttpFailure_ReturnsApplicationNotFound is not a valid test (errors mapping doesn't work as in the prev incarnation of Share)

digital_identity_client.go
```diff
@@ -63,11 +63,11 @@ func (client *DigitalIdentityClient) GetSdkID() string {
 }
 
 // CreateShareSession creates a sharing session to initiate a sharing process based on a policy
-func (client *DigitalIdentityClient) CreateShareSession(shareSession digitalidentity.ShareSessionRequest) (share digitalidentity.ShareSession, err error) {
+func (client *DigitalIdentityClient) CreateShareSession(shareSession *digitalidentity.ShareSessionRequest) (share digitalidentity.ShareSession, err error) {
```
fofiuiancu, line 66: params still need renaming

digital_identity_client_test.go
```diff
@@ -92,3 +91,32 @@ func TestDigitalIDClient_HttpFailure_ReturnsApplicationNotFound(t *testing.T) {
 	})
 	assert.Check(t, !temporary || !tempError.Temporary())
 }
+
+func ExampleGetSession() {
+	key, err := os.ReadFile("./test/test-key.pem")
+
+	mockSessionID := "SOME_SESSION_ID"
+	//mockClientSdkId := "SOME_CLIENT_SDK_ID"
+	//mockApiUrl := "https://example.com/api"
```
fofiuiancu, line 100: leftovers?

digital_identity_client_test.go
```diff
@@ -92,3 +91,32 @@ func TestDigitalIDClient_HttpFailure_ReturnsApplicationNotFound(t *testing.T) {
...
+	client, err := NewDigitalIdentityClient("some-sdk-id", key)
+
+	client.HTTPClient = &mockHTTPClient{
+		do: func(*http.Request) (*http.Response, error) {
+			return &http.Response{
+				StatusCode: 201,
```
fofiuiancu, line 106:
```suggestion
				StatusCode: 200,
```

digital_identity_client_test.go
```diff
@@ -92,3 +91,32 @@ func TestDigitalIDClient_HttpFailure_ReturnsApplicationNotFound(t *testing.T) {
...
+				Body:       io.NopCloser(strings.NewReader(`{"id":"SOME_ID","status":"SOME_STATUS","expiry":"SOME_EXPIRY","created":"SOME_CREATED","updated":"SOME_UPDATED","qrCode":{"id":"SOME_QRCODE_ID"},"receipt":{"id":"SOME_RECEIPT_ID"}}`)),
+			}, nil
+		},
+	}
+
+	result, err := client.GetSession(mockSessionID)
+	if err != nil {
+		return
+	}
+	fmt.Printf("Status code: %s", result.Status)
```
fofiuiancu, line 116: we could test more that the status

digitalidentity/share_session_builder_test.go
```diff
@@ -20,17 +20,17 @@ func ExampleShareSessionBuilder() {
 	}
 
 	fmt.Println(string(data))
-	// Output: {"policy":{"wanted":[],"wanted_auth_types":[],"wanted_remember_me":false},"extensions":[],"redirectUri":"","notification":{"url":"","method":"","verifyTls":null,"headers":null}}
```
fofiuiancu, line 23: does it matter for the PIA is these are null or empty arrays?

digital_identity_client_test.go
```diff
@@ -54,25 +56,22 @@ func TestYotiClient_CreateShareSession(t *testing.T) {
...
-	result, err := client.CreateShareSession(session)
+	result, err := client.CreateShareSession(&session)
 
-	b, err := json.Marshal(result)
-	if err != nil {
-		fmt.Println(err)
-		return
+	out, err2 := json.Marshal(result)
+	if err2 == nil {
+		fmt.Printf("OUR: %s", out)
```
fofiuiancu, line 66: leftover?

yotierror/response.go
```diff
@@ -38,6 +38,8 @@ type Error struct {
 	message  string
 	Err      error
 	Response *http.Response
+	id       string
```
fofiuiancu, line 41: just adding these new fields here won't work.
If you look here https://github.com/getyoti/yoti-go-sdk/blob/master/yotierror/response.go#L61C15-L137
You will see that the prev error format operated on the assumption that there is one HTTP status code that can be uniquely mapped to an error message.
That is not the case anymore as we how we can't rely on that mapping (and shouldn't do it in our code anymore). We should now just parse the body of response to get the error's details.
It will be simpler than before. 
We  probably need a new struct that describes the new error format, and prob it's best to have our own version of https://github.com/getyoti/yoti-go-sdk/tree/master/requests adequate for V2

fofiuiancu (changes requested):

yotierror/response.go
fofiuiancu: eg. bc the current code is still trying to do that mapping. instead of a properly formatted error you'll get an error with:

```
id = ""
status = 0
message = "401: unknown HTTP error - {"id":"8f6a9dfe72128de20909af0d476769b6","status":401,"error":"INVALID_REQUEST_SIGNATURE","message":"Invalid request signature"}"
```

digitalidentity/service.go
klaidas: i don't see the changes for this from the newest commits @mehmet-yoti - feel free to ping me on slack if you wanna chat about it abit quicker

klaidas:
agree with some of the comments on the tests that iancu mentioned, for me the main thing is that this still needs to clean up the error handling for the new API responses before merging

fofiuiancu:
> > > Not covered by this PR but should have been covered by the previous PR: The ShareSessionRequest should have the optional fields as pointers: (eg policy is mandatory in Share V2 so it shouldn't be a pointer)
> > > Please see the optional ones here: https://lampkicking.atlassian.net/wiki/spaces/DAS/pages/3900801110/Share+V2+Payload+documentation
> > 
> > 
> > this should also be addressed
> 
> ShareSessionNotification should also be optional and the Builder should take care of initialising the optionals with defaults where is the case (I a not sure why the extensions is populated in the Builder with an empty array, but haven't investigated it in depth)

It's not clear from the documentation is we should  initialise the optionals or if Connect does that if they are not shared.

https://lampkicking.atlassian.net/wiki/spaces/DAS/pages/3900801110/Share+V2+Payload+documentation

digitalidentity/service.go
```diff
@@ -28,14 +28,16 @@ func CreateShareSession(httpClient requests.HttpClient, shareSessionRequest *Sha
 		HTTPMethod: http.MethodPost,
 		BaseURL:    apiUrl,
 		Endpoint:   endpoint,
-		Headers:    requests.AuthHeader(clientSdkId),
+		Headers:    requests.AuthHeader(clientSdkId, &key.PublicKey),
```
fofiuiancu, line 31: it seems that we don't need the public key here; do we need the clientSDKId as we're also adding it a few rows later as a Param?

digitalidentity/share_session_builder.go
```diff
@@ -33,7 +33,7 @@ func (builder *ShareSessionRequestBuilder) WithExtension(extension interface{})
 
 // WithNotification sets the callback URL
 func (builder *ShareSessionRequestBuilder) WithNotification(notification ShareSessionNotification) *ShareSessionRequestBuilder {
-	builder.shareSessionRequest.shareSessionNotification = notification
+	builder.shareSessionRequest.shareSessionNotification = &notification
```
fofiuiancu, line 36: we should pass notification as a *ShareSessionNotification instead

digitalidentity/share_session_notification_builder_test.go
```diff
@@ -18,7 +18,7 @@ func ExampleShareSessionNotificationBuilder() {
 	}
 
 	fmt.Println(string(data))
-	// Output: {"url":"","method":"","verifyTls":null,"headers":null}
+	// Output: {"url":""}
```
fofiuiancu, line 21: maybe here is the point the defaults should have been set?

requests/signed_message.go
```diff
@@ -42,9 +42,16 @@ func JSONHeaders() map[string][]string {
 }
 
 // AuthHeader is a header prototype including the App/SDK ID
-func AuthHeader(clientSdkId string) map[string][]string {
+func AuthHeader(clientSdkId string, key *rsa.PublicKey) map[string][]string {
 	return map[string][]string{
 		"X-Yoti-Auth-Id": {clientSdkId},
+		/*"X-Yoti-Auth-Key": {
```
fofiuiancu, line 48: leftover

sonar-project.properties
```diff
@@ -11,4 +11,4 @@ sonar.go.coverage.reportPaths = coverage.out
 sonar.go.tests.reportPaths = sonar-report.json
 sonar.tests = .
 sonar.test.inclusions = **/*_test.go
-sonar.coverage.exclusions = test/**/*,_examples/**/*
+sonar.coverage.exclusions = test/**/*,_examples/**/*,src/digitalidentity/**
```
fofiuiancu, line 14: I am not sure this is the right approach. How did the other SDKs solve it?

yotierror/response.go
fofiuiancu: these fields have to be removed as we will have a new struct

fofiuiancu (changes requested):
see comments

digitalidentity/service.go
mehmet-yoti: As read from documentation backend needs this header.

yotierror/response.go
```diff
@@ -65,8 +65,8 @@ func formatResponseMessage(response *http.Response, httpErrorMessages ...map[int
 		return defaultMessage
 	}
 
-	body, _ := io.ReadAll(response.Body)
-	response.Body = io.NopCloser(bytes.NewBuffer(body))
+	body, _ := ioutil.ReadAll(response.Body)
```
klaidas, line 68: `ioutil` is deprecated, `io` was fine, these errors should've really been handled before though..

digitalidentity/share_session_notification_builder.go
```diff
@@ -49,10 +49,10 @@ func (b *ShareSessionNotificationBuilder) Build() (ShareSessionNotification, err
...
 func (a *ShareSessionNotification) MarshalJSON() ([]byte, error) {
 	return json.Marshal(&struct {
-		Url       string              `json:"url"`
-		Method    string              `json:"method"`
-		VerifyTls *bool               `json:"verifyTls"`
-		Headers   map[string][]string `json:"headers"`
+		Url       string               `json:"url"`
+		Method    *string              `json:"method,omitempty"`
+		VerifyTls *bool                `json:"verifyTls,omitempty"`
+		Headers   *map[string][]string `json:"headers,omitempty"`
```
klaidas, line 55:
```suggestion
		Headers   map[string][]string `json:"headers,omitempty"`
```

does not need to be pointer to map, maps are inherently pointers and can be nullable already

digitalidentity/share_session_notification_builder.go
```diff
@@ -7,9 +7,9 @@ import (
 // ShareSessionNotification specifies the session notification configuration.
 type ShareSessionNotification struct {
 	url       string
-	method    string
+	method    *string
 	verifyTLS *bool
-	headers   map[string][]string
+	headers   *map[string][]string
```
klaidas, line 12: same thing, does not need to be pointer here

digitalidentity/share_session_builder.go
```diff
@@ -4,76 +4,69 @@ import (
...
-type ShareSessionBuilder struct {
-	shareSession ShareSession
-	err          error
+// ShareSessionRequestBuilder builds a session
+type ShareSessionRequestBuilder struct {
+	shareSessionRequest ShareSessionRequest
+	err                 error
 }
 
 // ShareSession represents a sharesession
```
klaidas, line 13:
```suggestion
// ShareSessionRequest represents a sharesession
```

digitalidentity/service_test.go
```diff
@@ -104,12 +82,34 @@ func createShareUrlWithErrorResponse(statusCode int, responseBody string) (share
...
+			return &http.Response{
+				StatusCode: 201,
+				Body:       io.NopCloser(strings.NewReader(`{"id":"SOME_ID","status":"SOME_STATUS","expiry":"SOME_EXPIRY","created":"SOME_CREATED","updated":"SOME_UPDATED","qrCode":{"id":"SOME_QRCODE_ID"},"receipt":{"id":"SOME_RECEIPT_ID"}}`)),
+			}, nil
+		},
+	}
+
+	result, err := GetShareSession(client, mockSessionID, mockClientSdkId, mockApiUrl, key)
+	if err != nil {
+		return
```
klaidas, line 111: shouldn't this be assert no error?

digitalidentity/yotierror/signed_requests.go
```diff
@@ -0,0 +1,8 @@
+package yotierror
+
+const (
```
klaidas, line 3: these aren't used

digitalidentity/requests/signed_message.go
```diff
@@ -0,0 +1,222 @@
+package requests
```
klaidas, line 1: what was the reason for copying this entire package? it's better to have 1 copy of common code so we can maintain it easier and it won't bite us in the ass if we make a fix to one but forget the copy. 

if the whole reason was because we need it to use the new `yotierror` response to error parsing, we can just version the `Execute` function like `Execute2` or something (or if the name is important like it will be used directly by integrators then we could just have this function by itself inside the `digitalidentity` package instead) and have that do the new error handling

digitalidentity/service.go
fofiuiancu: It seems that we don't need it the Params (but we need it in the Headers)

sonar-project.properties
fofiuiancu: We should only exclude digitalidentity only from code duplication checks (so use sonar.cpd.exclusions)

digitalidentity/yotierror/signed_requests.go
fofiuiancu: True, but this is a package meant to be used by integrators so I;d like to keep them in if we believe they can add value as consts to compare against.

digitalidentity/requests/signed_message.go
fofiuiancu: Good points, but, as we plan to retire V1 I thought this version is preferable. That's the same reason the choice was taken to duplicate the Policy code. See the notes here: https://lampkicking.atlassian.net/browse/SDK-2230.

_examples/idv/handlers.session.go
```diff
@@ -63,6 +63,7 @@ func pageFromSessionSpec(c *gin.Context, sessionSpec *create.SessionSpecificatio
 		return
 	}
 	createSessionResult, err = client.CreateSession(sessionSpec)
+
```
fofiuiancu, line 66: we can remove this

digitalidentity/yotierror/response.go
```diff
@@ -0,0 +1,56 @@
...
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+)
+
+var (
+	defaultUnknownErrorCodeConst    = "UNKNOWN_ERROR"
+	defaultUnknownErrorMessageConst = "unknown HTTP error"
```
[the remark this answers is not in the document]
fofiuiancu: Just kept the convention from the V1 SDK

fofiuiancu (commented):
Only some minor things left, for me the main one is "use sonar.cpd.exclusions"

digitalidentity/yotierror/response.go
```diff
@@ -0,0 +1,56 @@
...
+)
+
+// Error indicates errors related to the Yoti API.
+type Error struct {
+	Id        string `json:"id"`
```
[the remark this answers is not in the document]
mehmet-yoti: I have checked it again it has id in response like below

{
  "id": "TCR1yNPVTECiGxyH95ix4g",
  "error": "SOME_ERROR_CODE",
  "status": 400,
  "message": "lorem ipsum ..."
}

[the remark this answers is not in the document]
klaidas: yh it's just not documented, ok i'll resolve

digitalidentity/yotierror/signed_requests.go
klaidas: well they would add value if they were used, if they're not used why would they need to compare against them?

digitalidentity/yotierror/response.go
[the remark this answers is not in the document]
klaidas: that's not a good convention to follow 😆

klaidas (approved):

fofiuiancu (approved):

PR #291, opened 2023-10-31
SDK-2379: added support for Advanced Identity Profile Requirements for idv and share v1

klaidas:
https://lampkicking.atlassian.net/browse/WAS-611

covers https://lampkicking.atlassian.net/browse/SDK-2379 and has some more details within that epic


klaidas:
Since @fofiuiancu started this work and i just took it over when it was pretty much almost done, just reviewed it and fixed and added some things i think only 1 reviewer other than us two is needed.

mehmet-yoti (approved):
Looks good.

PR #303, opened 2024-01-09
Sdk 2422 go fix code smells

mehmet-yoti:


digitalidentity/service.go
```diff
@@ -20,6 +20,9 @@ const identitySessionQrCodeCreation = "/v2/sessions/%s/qr-codes"
 const identitySessionQrCodeRetrieval = "/v2/qr-codes/%s"
 const identitySessionReceiptRetrieval = "/v2/receipts/%s"
 const identitySessionReceiptKeyRetrieval = "/v2/wrapped-item-keys/%s"
+const errorFailedToGetSignedReceipt = "failed to get signed request: %v"
```
fofiuiancu, line 23:
```suggestion
const errorFailedToGetSignedRequest = "failed to get signed request: %v"
```

requests/signed_message_test.go
```diff
@@ -167,3 +167,35 @@ func TestSignedRequest_WithPemFile_NotRSAKeyShouldError(t *testing.T) {
 -----END RSA PUBLIC KEY-----`))
 	assert.ErrorContains(t, msg.Error, "not an RSA Private Key")
 }
+
+// TestAuthHeader tests the AuthHeader function.
```
fofiuiancu, line 171: The AuthHeader function in request/signed_message.go is not used so let's remove the function there, and move this test so it tests the same function we have here https://github.com/getyoti/yoti-go-sdk/blob/SDK-2422-go-fix-code-smells/digitalidentity/requests/signed_message.go#L45

fofiuiancu (changes requested):

fofiuiancu (approved):

PR #314, opened 2025-05-06
SDK-2356: Added ErrorReason to SharedReceiptResponse

fofiuiancu:
Expanded SharedReceiptResponse to include details about the error:

**Before:**

```go
type SharedReceiptResponse struct {
	ID                 string
	SessionID          string
	RememberMeID       string
	ParentRememberMeID string
	Timestamp          string
	Error              string
	UserContent        UserContent
	ApplicationContent ApplicationContent
}
```

**After:**

```go
type SharedReceiptResponse struct {
	ID                 string
	SessionID          string
	RememberMeID       string
	ParentRememberMeID string
	Timestamp          string
	Error              string
	ErrorReason        ErrorReason
	UserContent        UserContent
	ApplicationContent ApplicationContent
}

type ErrorReason struct {
	RequirementsNotMetDetails []RequirementsNotMetDetail `json:"requirements_not_met_details"`
}

type RequirementsNotMetDetail struct {
	Details                string `json:"details"`
	AuditId                string `json:"audit_id"`
	FailureType            string `json:"failure_type"`
	DocumentType           string `json:"document_type"`
	DocumentCountryIsoCode string `json:"document_country_iso_code"`
}
```

### DIFF
--- a/digitalidentity/receipt.go
+++ b/digitalidentity/receipt.go
@@ -5,15 +5,26 @@ type Content struct {
 	ExtraData []byte `json:"extraData"`
 }
 
+type ErrorReason struct {
+	RequirementsNotMetDetails []struct {
+		Details                string `json:"details"`
+		AuditId                string `json:"audit_id"`
+		FailureType            string `json:"failure_type"`
+		DocumentType           string `json:"document_type"`
+		DocumentCountryIsoCode string `json:"document_country_iso_code"`
+	} `json:"requirements_not_met_details"`
+}
+
 type ReceiptResponse struct {
-	ID                 string   `json:"id"`
-	SessionID          string   `json:"sessionId"`
-	Timestamp          string   `json:"timestamp"`
-	RememberMeID       string   `json:"rememberMeId,omitempty"`
-	ParentRememberMeID string   `json:"parentRememberMeId,omitempty"`
-	Content            *Content `json:"content,omitempty"`
-	OtherPartyContent  *Content `json:"otherPartyContent,omitempty"`
-	WrappedItemKeyId   string   `json:"wrappedItemKeyId"`
-	WrappedKey         []byte   `json:"wrappedKey"`
-	Error              string   `json:"error"`
+	ID                 string      `json:"id"`
+	SessionID          string      `json:"sessionId"`
+	Timestamp          string      `json:"timestamp"`
+	RememberMeID       string      `json:"rememberMeId,omitempty"`
+	ParentRememberMeID string      `json:"parentRememberMeId,omitempty"`
+	Content            *Content    `json:"content,omitempty"`
+	OtherPartyContent  *Content    `json:"otherPartyContent,omitempty"`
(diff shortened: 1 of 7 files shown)


digitalidentity/receipt.go
```diff
@@ -5,15 +5,26 @@ type Content struct {
 	ExtraData []byte `json:"extraData"`
 }
 
+type ErrorReason struct {
```
klaidas, line 8: better to just keep this struct for the singular object and just make arrays of this struct where needed, that way we can reference this in code and pass it around nicely in functions

digitalidentity/receipt.go
```diff
@@ -5,15 +5,26 @@ type Content struct {
...
+	RequirementsNotMetDetails []struct {
+		Details                string `json:"details"`
+		AuditId                string `json:"audit_id"`
+		FailureType            string `json:"failure_type"`
+		DocumentType           string `json:"document_type"`
+		DocumentCountryIsoCode string `json:"document_country_iso_code"`
+	} `json:"requirements_not_met_details"`
```
klaidas, line 15: is the connect API actually mixing snake and camel case or is this just a bug here?

digitalidentity/service.go
```diff
@@ -226,6 +226,18 @@ func GetShareReceipt(httpClient requests.HttpClient, receiptId string, clientSdk
...
+			ID:                 receiptResponse.ID,
+			SessionID:          receiptResponse.SessionID,
+			RememberMeID:       receiptResponse.RememberMeID,
+			ParentRememberMeID: receiptResponse.ParentRememberMeID,
+			Timestamp:          receiptResponse.Timestamp,
+			Error:              receiptResponse.Error,
+			ErrorReason:        receiptResponse.ErrorReason,
+		}, nil
+	}
+
```
klaidas, line 240: we can remove `Error: receiptResponse.Error,` at the end of this function then in the success receipt path

digitalidentity/service.go
```diff
@@ -226,6 +226,18 @@ func GetShareReceipt(httpClient requests.HttpClient, receiptId string, clientSdk
 		return receipt, fmt.Errorf("failed to get receipt: %v", err)
 	}
 
+	if receiptResponse.Error != "" {
+		return SharedReceiptResponse{
+			ID:                 receiptResponse.ID,
+			SessionID:          receiptResponse.SessionID,
+			RememberMeID:       receiptResponse.RememberMeID,
+			ParentRememberMeID: receiptResponse.ParentRememberMeID,
```
klaidas, line 234: might not want to give a false impression that this is set in the failure receipt
```suggestion
```

digitalidentity/receipt.go
fofiuiancu: I moved RequirementsNotMetDetail to it's own type but for the rest we should mirror as exactly as possible the API (no surprises) and follow the same structure that was used in other SDKs for consistency.

klaidas (approved):

mehmet-yoti (approved):

PR #338, opened 2026-06-05
SDK-2777: Go - Add support for new capture_type property on Static Liveness resources - go

mehmet-yoti:
## Summary
Adds support for the new `capture_type` property on `StaticLivenessResourceResponse` as part of the Yoti Doc Scan Go SDK. The `capture_type` field (e.g. `"CAMERA"`) is now deserialized from the session retrieval API response, allowing callers to determine how a static liveness image was captured. This change is additive and backward-compatible — resources without `capture_type` in the response will have an empty string value.

## Changes
- **`docscan/session/retrieve/static_liveness_resource_response.go`** — Added `CaptureType string` field with JSON tag `capture_type` to `StaticLivenessResourceResponse`.
- **`test/fixtures/resource-container-static.json`** — Added `"capture_type": "CAMERA"` to the first fixture entry to represent a resource that includes the new field; the remaining two entries are left without it to cover the optional/absent case.
- **`docscan/session/retrieve/resource_container_test.go`** — Extended `TestStaticLivenessResourceResponse_UnmarshalJSON` to assert `CaptureType` is correctly populated (`"CAMERA"`) or empty string when absent. Added `TestZoomLivenessResourceResponse_HasNoCaptureType` to confirm `ZoomLivenessResourceResponse` is unaffected by this change and its existing fields still deserialize correctly.

## QA Test Steps

1. **Setup** — Clone the repo, check out this branch, and run `go test ./docscan/session/retrieve/...` to confirm all tests pass.

2. **Happy path — `capture_type` present**
   - Construct or mock a session retrieval response JSON where a `liveness_capture` entry has `"liveness_type": "STATIC"` and `"capture_type": "CAMERA"`.
   - Unmarshal into `ResourceContainer` and call `StaticLivenessResources()`.
   - Assert `CaptureType == "CAMERA"` on the matching resource.

3. **Backward-compatible — `capture_type` absent**
   - Use a session retrieval response JSON where a `liveness_capture` entry has `"liveness_type": "STATIC"` but no `capture_type` key.
   - Unmarshal and call `StaticLivenessResources()`.
   - Assert `CaptureType == ""` (zero value, no error).

4. **Zoom liveness unaffected**
   - Use the existing `resource-container.json` fixture (Zoom resources).
   - Unmarshal and call `ZoomLivenessResources()`.
   - Assert the Zoom resources have no `CaptureType` field and their existing fields (`Frames`, `FaceMap`) still deserialize correctly.

5. **Regression — existing static liveness fields**
   - Confirm `Image`, `LivenessType`, `ID`, `Tasks`, and `Source` fields on `StaticLivenessResourceResponse` are still populated correctly after this change.

6. **Full test suite** — Run `go test ./...` from the repo root and confirm no failures or regressions.

## Notes
- `CaptureType` is an optional field on the API side; the SDK represents absence as an empty string (Go zero value for `string`). Callers should check `CaptureType != ""` before using the value if they need to distinguish "not provided" from a blank value.
- This change only extends `StaticLivenessResourceResponse`; `ZoomLivenessResourceResponse` does not and should not have this field.
- No changes to request/create-session payloads — this is retrieval-side only.

---

*Related Jira:* SDK-2777
*Auto-generated by n8n + Claude CLI*

### DIFF
--- a/docscan/session/retrieve/resource_container_test.go
+++ b/docscan/session/retrieve/resource_container_test.go
@@ -34,6 +34,28 @@ func TestStaticLivenessResourceResponse_UnmarshalJSON(t *testing.T) {
 
 	assert.Equal(t, 3, len(result.LivenessCapture))
 	assert.Equal(t, "STATIC", result.LivenessCapture[0].LivenessType)
+
+	staticResources := result.StaticLivenessResources()
+	assert.Equal(t, 3, len(staticResources))
+	assert.Equal(t, "CAMERA", staticResources[0].CaptureType)
+	assert.Equal(t, "", staticResources[1].CaptureType)
+	assert.Equal(t, "", staticResources[2].CaptureType)
+}
+
+func TestZoomLivenessResourceResponse_HasNoCaptureType(t *testing.T) {
+	bytes, err := file.ReadFile("../../../test/fixtures/resource-container.json")
+	assert.NilError(t, err)
+
+	var result ResourceContainer
+	err = json.Unmarshal(bytes, &result)
+	assert.NilError(t, err)
+
+	zoomResources := result.ZoomLivenessResources()
+	assert.Equal(t, 1, len(zoomResources))
+	// ZoomLivenessResourceResponse does not have a CaptureType field
+	// Verify it compiles and returns the expected zoom-specific fields instead
+	assert.Equal(t, "IMAGE", zoomResources[0].Frames[0].Media.Type)
+	assert.Equal(t, "BINARY", zoomResources[0].FaceMap.Media.Type)
 }
 
 func TestLivenessResourceResponse_UnmarshalJSON_Invalid(t *testing.T) {
--- a/docscan/session/retrieve/resource_container_test.go
+++ b/docscan/session/retrieve/resource_container_test.go
@@ -42,22 +42,6 @@ func TestStaticLivenessResourceResponse_UnmarshalJSON(t *testing.T) {
 	assert.Equal(t, "", staticResources[2].CaptureType)
 }
 
-func TestZoomLivenessResourceResponse_HasNoCaptureType(t *testing.T) {
-	bytes, err := file.ReadFile("../../../test/fixtures/resource-container.json")
-	assert.NilError(t, err)
-
-	var result ResourceContainer
-	err = json.Unmarshal(bytes, &result)
-	assert.NilError(t, err)
-
-	zoomResources := result.ZoomLivenessResources()
-	assert.Equal(t, 1, len(zoomResources))
-	// ZoomLivenessResourceResponse does not have a CaptureType field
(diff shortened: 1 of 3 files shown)


mehmet-yoti:
## 🤖 Claude Code Review

## Code Review Findings

### Critical
None

### Major
None

### Minor

- **`docscan/session/retrieve/static_liveness_resource_response.go:7`** — `CaptureType` is modeled as a plain `string`, whereas every other optional-string field in this package that may be absent uses pointers so the caller can distinguish "absent" from "empty". Because `capture_type` is genuinely optional in the API (the fixture at `test/fixtures/resource-container-static.json:34` and `:50` omits it), consumers cannot tell a missing value from a server-sent empty string. If that distinction matters to API users, prefer `CaptureType *string \`json:"capture_type,omitempty"\``. If the team's convention is that simple optional strings stay as `string`, this is fine as-is — flagging for an explicit decision.

### Nit

- **`docscan/session/retrieve/resource_container.go:25-27`** — Pre-existing copy-paste in the doc comment for `StaticLivenessResources`: it reads `// ZoomLivenessResources ... only the "Zoom" liveness resources` but documents the Static method. Not introduced by this change, but it sits directly adjacent to the new `CaptureType` work and is a trivial fix: update to `// StaticLivenessResources filters the liveness resources, returning only the "Static" liveness resources`.

- **`docscan/session/retrieve/resource_container_test.go:45-59`** — `TestZoomLivenessResourceResponse_HasNoCaptureType` does not actually assert anything about `CaptureType`; it re-asserts the same `Frames`/`FaceMap` values already covered by the existing Zoom unmarshalling test. The comment explains the intent is a compile-time guarantee that Zoom has no `CaptureType` field, but a passing-by-default test gives a false sense of coverage and duplicates existing assertions. Consider dropping this test or renaming it to clarify it only verifies Zoom unmarshalling is unaffected by the new Static field.

- **`docscan/session/retrieve/static_liveness_resource_response.go:7`** — Consider adding `,omitempty` to the JSON tag (`json:"capture_type,omitempty"`) for symmetry with marshalling expectations, though since this struct is response-only (unmarshal path), the practical impact is negligible.

### What's done well
- The core change is minimal and idiomatic: adding the field to the struct lets `encoding/json` populate it automatically via the existing `UnmarshalJSON` dispatch, with no changes needed to the unmarshalling logic.
- Test coverage correctly exercises both the populated (`"CAMERA"`) and absent (`""`) cases against an updated fixture — exactly the right edge cases to verify for an optional field.
- The fixture update cleanly adds `capture_type` to only the first of three resources, deliberately preserving the "missing field" scenario for the other two.

J-Stead (approved):
LGTM

But it doesn't seem necessary to check that zoom liveness does not have a capture type in the test

saurabh-yoti (approved):

PR #339, opened 2026-06-16
SDK-2788: Go - Add support for retrieving the extraction_image_ids field from the IDV pages - go

mehmet-yoti:
## Summary
Adds support for the `extraction_image_ids` field on `PageResponse` in the IDV (Identity Verification) session retrieval API. This field returns a list of UUID strings identifying the images extracted from a document page during processing. The change extends the existing `PageResponse` struct with the new field and updates test fixtures to reflect the new API shape.

## Changes
- **`docscan/session/retrieve/page_response.go`** — Added `ExtractionImageIds []string` field with `json:"extraction_image_ids"` tag to `PageResponse`.
- **`docscan/session/retrieve/page_response_test.go`** — Added unit tests covering: populated array, empty array, `null` value, absent field, and JSON round-trip serialization for `ExtractionImageIds`.
- **`test/fixtures/watchlist_advanced_ca_profile_custom.json`** — Updated fixture pages to include `extraction_image_ids` with two UUIDs on one page and an empty array on the other, reflecting real API variance.
- **`test/fixtures/watchlist_screening.json`** — Updated fixture page to include a single UUID in `extraction_image_ids`, aligning the fixture with the updated API contract.

## QA Test Steps

1. **Setup**: Check out branch `websdk-auto/SDK-2788-go-add-support-for-retrieving-the-extraction-image-ids-field-from-the-idv-pages` and run `go test ./docscan/session/retrieve/...` — all tests should pass.

2. **Happy path — populated IDs**: Confirm `TestPageResponse_ExtractionImageIds_WithValues` passes: `PageResponse` unmarshals JSON with `extraction_image_ids: ["uuid-1", "uuid-2"]` and returns a slice of length 2 with correct values.

3. **Edge case — empty array**: Confirm `TestPageResponse_ExtractionImageIds_EmptyArray` passes: JSON with `extraction_image_ids: []` results in a zero-length slice, not `nil`.

4. **Edge case — null field**: Confirm `TestPageResponse_ExtractionImageIds_NullValue` passes: JSON with `extraction_image_ids: null` results in a zero-length slice.

5. **Edge case — field absent**: Confirm `TestPageResponse_ExtractionImageIds_FieldAbsent` passes: JSON without the field at all results in a zero-length slice (zero value for `[]string`).

6. **Round-trip serialization**: Confirm `TestPageResponse_ExtractionImageIds_RoundTrip` passes: marshal then unmarshal a `PageResponse` with IDs and verify the values survive intact.

7. **Fixture-driven integration tests**: Run the full test suite (`go test ./...`) and confirm tests that load `test/fixtures/watchlist_advanced_ca_profile_custom.json` and `test/fixtures/watchlist_screening.json` still pass with the updated fixture data.

8. **Regression**: Verify that existing `PageResponse` fields (`CaptureMethod`, `Media`, `Frames`) are still correctly populated when `extraction_image_ids` is present in the JSON.

## Notes
- `ExtractionImageIds` uses Go's zero value (`nil` slice) when absent or `null` in JSON; callers should use `len()` rather than a `nil` check to safely handle all cases.
- The fixture updates are non-breaking — they only add the new field to existing page objects, so no other fixture-consuming tests should regress.
- No getter method was added; the field is directly accessible on the struct, consistent with the pattern used by `CaptureMethod`, `Media`, and `Frames` in the same struct.

---

*Related Jira:* SDK-2788
*Auto-generated by n8n + Claude CLI*

### DIFF
--- a/docscan/session/retrieve/page_response.go
+++ b/docscan/session/retrieve/page_response.go
@@ -2,7 +2,8 @@ package retrieve
 
 // PageResponse represents information about an uploaded document Page
 type PageResponse struct {
-	CaptureMethod string           `json:"capture_method"`
-	Media         *MediaResponse   `json:"media"`
-	Frames        []*FrameResponse `json:"frames"`
+	CaptureMethod      string           `json:"capture_method"`
+	Media              *MediaResponse   `json:"media"`
+	Frames             []*FrameResponse `json:"frames"`
+	ExtractionImageIds []string         `json:"extraction_image_ids"`
 }
--- a/docscan/session/retrieve/page_response_test.go
+++ b/docscan/session/retrieve/page_response_test.go
@@ -1,0 +1,80 @@
+package retrieve
+
+import (
+	"encoding/json"
+	"testing"
+
+	"gotest.tools/v3/assert"
+)
+
+func TestPageResponse_ExtractionImageIds_WithValues(t *testing.T) {
+	jsonData := []byte(`{
+		"capture_method": "UPLOAD",
+		"extraction_image_ids": ["uuid-1", "uuid-2"]
+	}`)
+
+	var page PageResponse
+	err := json.Unmarshal(jsonData, &page)
+	assert.NilError(t, err)
+
+	assert.Equal(t, 2, len(page.ExtractionImageIds))
+	assert.Equal(t, "uuid-1", page.ExtractionImageIds[0])
+	assert.Equal(t, "uuid-2", page.ExtractionImageIds[1])
+}
+
+func TestPageResponse_ExtractionImageIds_EmptyArray(t *testing.T) {
+	jsonData := []byte(`{
+		"capture_method": "UPLOAD",
+		"extraction_image_ids": []
+	}`)
+
+	var page PageResponse
+	err := json.Unmarshal(jsonData, &page)
+	assert.NilError(t, err)
+
+	assert.Equal(t, 0, len(page.ExtractionImageIds))
+}
+
+func TestPageResponse_ExtractionImageIds_NullValue(t *testing.T) {
+	jsonData := []byte(`{
+		"capture_method": "UPLOAD",
+		"extraction_image_ids": null
+	}`)
+
+	var page PageResponse
+	err := json.Unmarshal(jsonData, &page)
+	assert.NilError(t, err)
+
+	assert.Equal(t, 0, len(page.ExtractionImageIds))
+}
+
+func TestPageResponse_ExtractionImageIds_FieldAbsent(t *testing.T) {
+	jsonData := []byte(`{
+		"capture_method": "UPLOAD"
+	}`)
+
+	var page PageResponse
+	err := json.Unmarshal(jsonData, &page)
+	assert.NilError(t, err)
+
+	assert.Equal(t, 0, len(page.ExtractionImageIds))
+}
+
+func TestPageResponse_ExtractionImageIds_RoundTrip(t *testing.T) {
+	ids := []string{"uuid-aaa", "uuid-bbb"}
+	page := PageResponse{
+		CaptureMethod:      "CAMERA",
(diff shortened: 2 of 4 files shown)


mehmet-yoti:
## 🤖 Claude Code Review

## Code Review Findings

### Summary
The change adds a single `ExtractionImageIds []string` field to `PageResponse` so the `extraction_image_ids` field returned by the IDV pages endpoint is deserialized. The implementation is minimal, idiomatic, and consistent with the surrounding package (plain public struct fields, `Id`/`Ids` naming matching `SubjectId`/`SessionId`/`AuditId`). Build passes and the new unit tests cover the meaningful unmarshalling cases. No correctness, security, or performance concerns.

### Critical
None

### Major
None

### Minor

- **`test/fixtures/watchlist_screening.json:39` and `test/fixtures/watchlist_advanced_ca_profile_custom.json:44,55`** — The fixtures were updated to include `extraction_image_ids`, but the tests that consume them (`get_session_result_test.go:123` `TestGetSessionResult_UnmarshalJSON_Watchlist` and `get_session_result_test.go:145` `TestGetSessionResult_UnmarshalJSON_Watchlist_Advanced_CA`) never assert on the field. As written, the fixture additions exercise no assertion path, so they don't actually protect against regressions in the full `GetSessionResult` -> pages deserialization chain (`PageResponse` is reached via `IDDocumentResourceResponse.Pages` / `SupplementaryDocumentResourceResponse.Pages`).
  - *Why it matters:* The dedicated `page_response_test.go` tests unmarshal `PageResponse` in isolation, which is good, but nothing verifies the field survives end-to-end through the real session-result parsing that these fixtures feed. Modifying fixtures without an assertion gives a false sense of coverage.
  - *Suggested fix:* Add an assertion in one of the watchlist tests reaching into a page, e.g.:
    ```go
    page := watchlistScreeningCheck.GeneratedProfile /* ...navigate to the resource with pages... */
    assert.Equal(t, "a1b2c3d4-0000-0000-0000-000000000001", page.ExtractionImageIds[0])
    ```
    Or, if the watchlist fixtures don't expose a navigable page path from the public API, revert the fixture edits and instead extend an existing IDV-document fixture/test (one that asserts on `.Pages`) so the change is verified through the real call path.

### Nit

- **`docscan/session/retrieve/page_response.go:8`** — Consider a brief doc comment on the new field, consistent with the documented `Pages` fields in `id_document_resource_response.go:16`. e.g. `// ExtractionImageIds are the IDs of images used for data extraction on this page`. The struct already has a type-level comment, so this is purely a polish item.

- **`docscan/session/retrieve/page_response_test.go:38` (`TestPageResponse_ExtractionImageIds_NullValue`)** — The test asserts `len == 0` for a JSON `null`, which is correct, but a `null` unmarshals to a `nil` slice while `[]` unmarshals to a non-nil empty slice. If callers ever distinguish nil-vs-empty (they shouldn't need to here), that nuance is untested. Optional: add `assert.Assert(t, page.ExtractionImageIds == nil)` in the null case and `!= nil` in the empty-array case to document the intended distinction. Not required given the field is read-only output.

### What's Done Well
- Field naming (`ExtractionImageIds`) and the plain public-field approach match the established convention in this package precisely — appropriately minimal, no getter or custom `UnmarshalJSON` needed.
- `page_response_test.go` is thorough for the unit it targets: values present, empty array, null, absent field, and a marshal/unmarshal round-trip — the four JSON edge cases that matter for a slice field are all covered.
- The change is correctly scoped to deserialization only, with no unnecessary API surface added.

### Overall
**Approved with Comments** — The code itself is ready to merge. The one worthwhile follow-up is adding an assertion so the modified fixtures actually verify the field end-to-end (or reverting those fixture edits if they aren't on an assertable path).

J-Stead (approved):
LGTM

saurabh-yoti (approved):

PR #344, opened 2026-07-02
SDK-2761: Go - Support handled_check_limit property for checks in IDV sandbox - go

mehmet-yoti:
## Summary

Adds support for the `handled_check_limit` property on IDV sandbox check-report builders. The field is serialised as `handled_check_limit` in JSON (using `omitempty` so it is absent when not set), matching the behaviour expected by the Yoti IDV sandbox API. All eight concrete check-report builder types expose a new public `WithHandledCheckLimit(int)` method.

## Changes

- **`docscan/sandbox/request/check/check.go`** — Added `HandledCheckLimit *int` (`json:"handled_check_limit,omitempty"`) to the base `check` struct and a `withHandledCheckLimit(limit int)` helper on `checkBuilder`.
- **`docscan/sandbox/request/check/document_authenticity_check.go`** — Exposed `WithHandledCheckLimit(int)` on `DocumentAuthenticityCheckBuilder`.
- **`docscan/sandbox/request/check/document_face_match_check.go`** — Exposed `WithHandledCheckLimit(int)` on `DocumentFaceMatchCheckBuilder`.
- **`docscan/sandbox/request/check/document_text_data_check.go`** — Exposed `WithHandledCheckLimit(int)` on `DocumentTextDataCheckBuilder`.
- **`docscan/sandbox/request/check/id_document_comparison_check.go`** — Exposed `WithHandledCheckLimit(int)` on `IDDocumentComparisonCheckBuilder`.
- **`docscan/sandbox/request/check/supplementary_document_text_data_check.go`** — Exposed `WithHandledCheckLimit(int)` on `SupplementaryDocumentTextDataCheckBuilder`.
- **`docscan/sandbox/request/check/third_party_identity_check.go`** — Exposed `WithHandledCheckLimit(int)` on `ThirdPartyIdentityCheckBuilder`.
- **`docscan/sandbox/request/check/zoom_liveness_check.go`** — Exposed `WithHandledCheckLimit(int)` on `ZoomLivenessCheckBuilder`.
- **`docscan/sandbox/request/check/static_liveness_check.go`** — Exposed `WithHandledCheckLimit(int)` on `StaticLivenessCheckBuilder`.
- **`docscan/sandbox/request/check/handled_check_limit_test.go`** — New test file with per-type tests confirming the field is included in JSON when set and omitted when not set, plus a runnable Example function.

## QA Test Steps

1. **Setup** — Clone the repo, check out this branch, and ensure Go tooling is available (`go version`).
2. **Unit tests** — Run `go test ./docscan/sandbox/request/check/...` and confirm all tests pass, including the new `TestHandledCheckLimit_*` cases.
3. **Happy path — field present** — Call any builder (e.g. `NewDocumentAuthenticityCheckBuilder()`) with `.WithHandledCheckLimit(3)`, marshal to JSON, and verify `"handled_check_limit":3` appears in the output.
4. **Happy path — field absent** — Call the same builder without `.WithHandledCheckLimit(...)`, marshal to JSON, and verify `handled_check_limit` is NOT present in the output.
5. **Edge case — zero value** — Call `.WithHandledCheckLimit(0)`, marshal, and verify `"handled_check_limit":0` IS present (zero is a valid limit, not omitted).
6. **All check types** — Repeat steps 3-4 for each of the eight builder types to ensure consistent behaviour.
7. **Regression** — Run the full test suite (`go test ./...`) to confirm no existing tests are broken.

## Notes

- `*int` is used instead of `int` so that the zero value (0) can be distinguished from "not set" while still allowing `omitempty` to elide the field when nil.
- No changes to non-sandbox (production) code paths; this is sandbox-only.
- Follow-up work may be needed in other SDK languages if the same property is required there.

---

*Related Jira:* SDK-2761
*Auto-generated by Claude dynamic workflow*

### DIFF
--- a/docscan/sandbox/request/check/check.go
+++ b/docscan/sandbox/request/check/check.go
@@ -5,12 +5,14 @@ import (
 )
 
 type check struct {
-	Result checkResult `json:"result"`
+	Result            checkResult `json:"result"`
+	HandledCheckLimit *int        `json:"handled_check_limit,omitempty"`
 }
 
 type checkBuilder struct {
-	recommendation *report.Recommendation
-	breakdowns     []*report.Breakdown
+	recommendation    *report.Recommendation
+	breakdowns        []*report.Breakdown
+	handledCheckLimit *int
 }
 
 type checkResult struct {
@@ -30,6 +32,10 @@ func (b *checkBuilder) withBreakdown(breakdown *report.Breakdown) {
 	b.breakdowns = append(b.breakdowns, breakdown)
 }
 
+func (b *checkBuilder) withHandledCheckLimit(limit int) {
+	b.handledCheckLimit = &limit
+}
+
 func (b *checkBuilder) build() *check {
 	return &check{
 		Result: checkResult{
@@ -38,5 +44,6 @@ func (b *checkBuilder) build() *check {
 				Breakdown:      b.breakdowns,
 			},
 		},
+		HandledCheckLimit: b.handledCheckLimit,
 	}
 }
--- a/docscan/sandbox/request/check/check.go
+++ b/docscan/sandbox/request/check/check.go
@@ -33,7 +33,8 @@ func (b *checkBuilder) withBreakdown(breakdown *report.Breakdown) {
 }
 
 func (b *checkBuilder) withHandledCheckLimit(limit int) {
-	b.handledCheckLimit = &limit
+	b.handledCheckLimit = new(int)
+	*b.handledCheckLimit = limit
 }
 
 func (b *checkBuilder) build() *check {
--- a/docscan/sandbox/request/check/document_authenticity_check.go
+++ b/docscan/sandbox/request/check/document_authenticity_check.go
@@ -38,6 +38,12 @@ func (b *DocumentAuthenticityCheckBuilder) WithDocumentFilter(filter *filter.Doc
 	return b
 }
 
+// WithHandledCheckLimit sets the number of times this check response is used before advancing to the next
+func (b *DocumentAuthenticityCheckBuilder) WithHandledCheckLimit(limit int) *DocumentAuthenticityCheckBuilder {
+	b.withHandledCheckLimit(limit)
+	return b
+}
+
 // Build creates a new DocumentAuthenticityCheck
 func (b *DocumentAuthenticityCheckBuilder) Build() (*DocumentAuthenticityCheck, error) {
 	return &DocumentAuthenticityCheck{
--- a/docscan/sandbox/request/check/document_face_match_check.go
+++ b/docscan/sandbox/request/check/document_face_match_check.go
@@ -38,6 +38,12 @@ func (b *DocumentFaceMatchCheckBuilder) WithDocumentFilter(filter *filter.Docume
 	return b
 }
 
+// WithHandledCheckLimit sets the number of times this check response is used before advancing to the next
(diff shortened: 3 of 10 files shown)


mehmet-yoti:
## 🤖 Claude Code Review

## Code Review Findings

### Critical
None

### Major
None

### Minor

**1. `assertHandledCheckLimitInJSON` uses `strings.Contains` — can produce false positives**

`docscan/sandbox/request/check/handled_check_limit_test.go`, line 214

The helper builds the expected string as `"handled_check_limit":%d` and uses `strings.Contains`. A limit of `2` matches any serialised value that *starts* with `2` (e.g. `20`, `25`), because `"handled_check_limit":2` is a substring of `"handled_check_limit":25`. This means the assertion could silently pass for the wrong value in any future test that uses multi-digit limits.

Suggested fix — unmarshal and compare numerically:

```go
func assertHandledCheckLimitInJSON(t *testing.T, v interface{}, limit int) {
    t.Helper()
    data, err := json.Marshal(v)
    if err != nil {
        t.Fatalf("unexpected error marshaling: %v", err)
    }
    var m map[string]json.RawMessage
    if err := json.Unmarshal(data, &m); err != nil {
        t.Fatalf("unexpected error unmarshaling: %v", err)
    }
    raw, ok := m["handled_check_limit"]
    if !ok {
        t.Errorf("expected JSON to contain handled_check_limit, got: %s", string(data))
        return
    }
    var got int
    if err := json.Unmarshal(raw, &got); err != nil {
        t.Fatalf("could not unmarshal handled_check_limit: %v", err)
    }
    if got != limit {
        t.Errorf("expected handled_check_limit=%d, got %d", limit, got)
    }
}
```

---

**2. No validation for zero or negative `handled_check_limit`**

`docscan/sandbox/request/check/check.go`, line 35–37

`omitempty` on a `*int` only omits the field when the pointer is `nil`. Calling `WithHandledCheckLimit(0)` stores a non-nil pointer to `0`, so the value `"handled_check_limit":0` is serialised and sent to the API, which is semantically meaningless (the response would never be used) and likely unintended.

Suggested fix — add a guard:

```go
func (b *checkBuilder) withHandledCheckLimit(limit int) {
    if limit <= 0 {
        return // or return an error from the public builder method
    }
    b.handledCheckLimit = &limit
}
```

Alternatively, change the public method signatures to return `error` so callers can detect the invalid input.

---

### Nit

**3. Example function name uses lowercase method name — not linked to exported API in godoc**

`docscan/sandbox/request/check/handled_check_limit_test.go`, line 14

The function is named `ExampleDocumentAuthenticityCheckBuilder_withHandledCheckLimit`. Go's documentation tool links an `ExampleT_M` function to method `T.M` only when the suffix (`M`) starts with an **uppercase** letter. Because `withHandledCheckLimit` starts with a lowercase `w`, the tool treats it as a free-form disambiguation string and will not associate this example with the exported `WithHandledCheckLimit` method in the package docs.

Suggested fix:

```go
// ExampleDocumentAuthenticityCheckBuilder_WithHandledCheckLimit shows handled_check_limit serialization
func ExampleDocumentAuthenticityCheckBuilder_WithHandledCheckLimit() {
```

J-Stead:
LGTM 

The false positive minor issue seems irrelevant as clients should not be setting the limit in double digits

saurabh-yoti (approved):

PR #351, opened 2026-09-10
feat(SDK-2476): Add tracked devices endpoints to IDV/DocScan client

mehmet-yoti:
## Summary

Adds support for the IDV tracked-devices endpoints to the Go DocScan/IDV client. Two new methods — \`GetTrackedDevices\` and \`DeleteTrackedDevices\` — allow callers to retrieve and delete device metadata events associated with a session. New response model structs map the JSON payloads returned by the API.

## Changes

- **\`docscan/client.go\`** — Added \`GetTrackedDevices(sessionID string) ([]*retrieve.TrackedDeviceResponse, error)\` (HTTP GET) and \`DeleteTrackedDevices(sessionID string) error\` (HTTP DELETE) methods on \`*Client\`.
- **\`docscan/endpoint.go\`** — Added \`getTrackedDevicesPath\` and \`deleteTrackedDevicesPath\` helpers that build the \`/sessions/{sessionId}/tracked-devices\` path.
- **\`docscan/session/retrieve/tracked_device_response.go\`** — New file defining \`TrackedDeviceResponse\` (event, resource ID, created timestamp, device) and \`TrackedDeviceDeviceResponse\` (IP address, country code, device/OS/browser metadata) model structs with JSON tags.
- **\`docscan/client_test.go\`** — 9 new unit tests covering: successful GET and DELETE responses, malformed JSON, non-2xx HTTP status codes, and missing/empty session ID edge cases.

## QA Test Steps

1. **Setup**
   - Clone the repository and checkout branch \`websdk-auto/SDK-2476-go-device-metadata-endpoints\`.
   - Ensure Go 1.18+ is installed.
   - Run \`go mod download\` to restore dependencies.

2. **Unit tests (no credentials required)**
   - Run \`go test ./docscan/... -v -run TestTrackedDevice\` and confirm all 9 new tests pass.
   - Run \`go test ./docscan/...\` to verify no regressions in existing test suite.

3. **Happy path (integration, requires IDV credentials)**
   - Initialise a \`DocScan\` client with a valid SDK ID and PEM key.
   - Create an IDV session and complete a user journey on a mobile device.
   - Call \`GetTrackedDevices(sessionID)\` and confirm a non-empty slice of \`TrackedDeviceResponse\` is returned with expected fields (\`Event\`, \`Created\`, \`Device\`).
   - Call \`DeleteTrackedDevices(sessionID)\` and confirm no error is returned.
   - Call \`GetTrackedDevices(sessionID)\` again and confirm an empty slice is returned.

4. **Edge cases**
   - Pass an empty string as \`sessionID\` — expect an error to be returned.
   - Simulate a 404 response (invalid session ID) — expect the error to surface from the HTTP layer.
   - Simulate a 500 response — expect the error to surface correctly.

5. **Regression**
   - Run the full test suite (\`go test ./...\`) and confirm all pre-existing tests still pass.

## Notes

- The DELETE endpoint shares the same URL path as the GET endpoint (\`/sessions/{sessionId}/tracked-devices\`), differentiated only by HTTP method.
- \`ResourceID\` on \`TrackedDeviceResponse\` is a pointer and is only populated for \`RESOURCE_CREATED\` events; all other device metadata fields on \`TrackedDeviceDeviceResponse\` are also optional pointers.
- No breaking changes are introduced; the new methods are purely additive.
- Follow-up: consider exposing the tracked-devices response types at the top-level package if consumers need them directly.

---

*Related Jira:* SDK-2476
*Auto-generated by Claude dynamic workflow*


mehmet-yoti:
## 🤖 Claude Code Review

## Code Review Findings

### Critical

None

### Major

**1. `docscan/client.go` lines 309–311 — `response.Body` not closed after `ReadAll`**

`GetTrackedDevices` reads from `response.Body` but never closes it. In long-running or high-throughput services this leaks the underlying TCP connection. Every other method in `client.go` has the same omission (so this is a pre-existing pattern), but the new code continues the problem. A `defer response.Body.Close()` immediately after receiving a non-error response is the standard Go fix:

```go
response, err = requests.Execute(c.HTTPClient, request, yotierror.DefaultHTTPErrorMessages)
if err != nil {
    return nil, err
}
defer response.Body.Close()   // add this
```

The same applies anywhere `io.ReadAll(response.Body)` is called without a matching close (lines 101, 136, 204, 274, 309, 442).

### Minor

**2. `docscan/session/retrieve/tracked_device_response.go` line 22 — typo: `ManufactureName` should be `ManufacturerName`**

`ManufactureName` is missing the trailing `r`. The JSON tag `manufacture_name` suggests this field name comes from the API contract; if the API uses `manufacturer_name`, both the tag and the field should be corrected. If the API really does use `manufacture_name`, at least the Go field name should match standard English spelling (`ManufacturerName`) while keeping the JSON tag as-is:

```go
// Current
ManufactureName  *string `json:"manufacture_name,omitempty"`

// Preferred (if API tag is correct)
ManufacturerName *string `json:"manufacture_name,omitempty"`
```

**3. `docscan/endpoint.go` lines 33–35 — `deleteTrackedDevicesPath` is redundant**

`deleteTrackedDevicesPath` is a one-liner that simply delegates to `getTrackedDevicesPath`. The existing `deleteSessionPath`/`deleteMediaPath` functions follow the same pattern, so it is consistent with the rest of the file. However, if the two paths ever diverge, the indirection helps; if they cannot diverge (same REST resource, different verb), the function adds no value. At minimum a comment explaining the intent would help future readers.

**4. `docscan/session/retrieve/tracked_device_response.go` line 15 — `Device` field is not `omitempty`**

`Device *TrackedDeviceDeviceResponse` is a pointer but lacks `omitempty`. If the API can return a `null` device (e.g. for some event types), this is fine. But if `null` is never expected, adding `,omitempty` keeps the struct self-consistent with the other nullable pointer fields:

```go
Device *TrackedDeviceDeviceResponse `json:"device,omitempty"`
```

### Nit

**5. `docscan/client_test.go` line 730 — 1024-bit RSA key in new tests**

Newly added tests generate 1 024-bit RSA keys (`rsa.GenerateKey(rand.Reader, 1024)`). This matches the rest of the test file, but 1 024-bit keys are below modern security minimums. Since these are test-only keys the risk is low, but 2048 bits is the recommended minimum and would be more consistent with any future lint/security-scanner rules.

**6. `docscan/client_test.go` — `Created` timestamp field not asserted in `TestClient_GetTrackedDevices`**

The JSON fixture includes `"created": "2024-01-15T10:30:00Z"` but the test does not assert `firstEvent.Created`. Adding an assertion would confirm the `time.Time` deserialization works correctly and prevent regressions if the field is ever accidentally removed.

